PR #2, opened 2015-06-09
Added session based client.  Cleaned up a few tests.

avinash240:
The session based client is faster than the default one, especially at high rates.  It's no longer paying the connection and breakdown penalty per socket and instead is connection pooling, which is great.  There are two tests, the geo tests.  They don't return back the unique flag that you send in, I suspect that that index type does not support unique.  However, I'm new to arangodb, so I don't know perhaps it's a bug in the server I'm running 2.5.5.  I also added the sparse and selectivity keys where needed.  I'm not sure if this breaks backwards compat with older servers, I'm sorry if it does, I just didn't want to hand back a broken test suite especially since you went through the trouble to build one.

### DIFF
--- a/arango/api.py
+++ b/arango/api.py
@@ -3,6 +3,7 @@
 import json
 
 from arango.clients.default import DefaultArangoClient
+from arango.clients.session import SessionArangoClient
 from arango.utils import is_string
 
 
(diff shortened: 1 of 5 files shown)


Joowani:
Hello avinash240,

Thanks for your interest and help on py-arango! I've created a new branch called "development". Could you redirect your pull request to it? Thanks.

avinash240:
Sure. No worries.

PR #39, opened 2017-04-26
[PERF] improve performance of large batch commits

patricklx:


joowani:
Thanks for the PR @patricklx!

PR #62, opened 2017-11-09
Add changes to make BatchExecution thread safe

dancollins34:
Simple, blocking implementation of BatchExecution.  This implementation uses locks to manage write control over the object.  However, the lock implementation in python 2 does not have a timeout option.  Therefore, one had to be created for python 2 based on Queue.Queue.  In python 3, threading.Lock is used.

### DIFF
--- a/arango/batch.py
+++ b/arango/batch.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, unicode_literals
 
 from uuid import uuid4
 
+from arango.lock import Lock
 from arango.collections.standard import Collection
 from arango.connection import Connection
 from arango.utils import HTTP_OK
@@ -28,13 +29,16 @@ class BatchExecution(Connection):
         so far are committed even if an exception is raised before existing
         out of the context (default: ``False``)
     :type commit_on_error: bool
-
-    .. warning::
-        Batch execution is currently an experimental feature and is not
-        thread-safe.
+    :param submit_timeout: the timeout to use for acquiring the lock necessary
+        to submit a batch.  Only relevant in multi-threaded contexts. In single
+        threaded contexts, acquiring this lock will never fail.  A value of
+        <= 0 means never timeout, a positive integer value indicates the number
+        of seconds to wait.
+    :type submit_timeout: int
     """
 
-    def __init__(self, connection, return_result=True, commit_on_error=False):
+    def __init__(self, connection, return_result=True, commit_on_error=False,
+                 submit_timeout=-1):
         super(BatchExecution, self).__init__(
             protocol=connection.protocol,
             host=connection.host,
@@ -51,6 +55,11 @@ class BatchExecution(Connection):
         self._requests = []  # The queue for requests
         self._handlers = []  # The queue for response handlers
         self._batch_jobs = []  # For tracking batch jobs
+
+        # For ensuring additions to all queues in the same order
+        self._lock = Lock()
+
+        self._batch_submit_timeout = submit_timeout
         self._aql = AQL(self)
         self._type = 'batch'
 
@@ -82,17 +91,21 @@ class BatchExecution(Connection):
         :type request: arango.request.Request
         :param handler: the response handler
         :type handler: callable
-        :returns: the batch job or None
-        :rtype: arango.batch.BatchJob
+        :returns: the :class:`arango.batch.BatchJob` or None
+        :rtype: :class:`arango.batch.BatchJob` | None
         """
-        self._requests.append(request)
-        self._handlers.append(handler)
 
-        if not self._return_result:
-            return None
-        batch_job = BatchJob()
-        self._batch_jobs.append(batch_job)
-        return batch_job
+        to_return = None
+
+        with self._lock:
+            self._requests.append(request)
+            self._handlers.append(handler)
+
+            if self._return_result:
+                to_return = BatchJob()
+                self._batch_jobs.append(to_return)
+
+        return to_return
 
     def commit(self):
         """Execute the queued API requests in a single HTTP call.
@@ -102,12 +115,23 @@ class BatchExecution(Connection):
         for later retrieval via its :func:`arango.batch.BatchJob.result`
         method
 
+        :returns: list of :class:`arango.batch.BatchJob` or None
+        :rtype: [arango.batch.BatchJob] | None
+
         :raises arango.exceptions.BatchExecuteError: if the batch request
             cannot be executed
         """
+
+        res = self._lock.acquire(timeout=self._batch_submit_timeout)
+
+        if not res:
+            raise BatchExecuteError("Unable to reaccquire lock within time "
+                                    "period. Some thread must be holding it.")
+
         try:
-            if not self._requests:
+            if len(self._requests) == 0:
                 return
+
             raw_data_list = []
             for content_id, request in enumerate(self._requests, start=1):
                 raw_data_list.append('--XXXsubpartXXX\r\n')
@@ -155,10 +179,13 @@ class BatchExecution(Connection):
                     job.update(status='error', result=err)
                 else:
                     job.update(status='done', result=result)
+
+            return self._batch_jobs
         finally:
             self._requests = []
             self._handlers = []
             self._batch_jobs = []
+            self._lock.release()
 
     def clear(self):
         """Clear the requests queue and discard pointers to batch jobs issued.
@@ -224,6 +251,7 @@ class BatchJob(object):
         self._id = uuid4()
         self._status = 'pending'
         self._result = None
+        self._lock = Lock()
 
     def __repr__(self):
         return '<ArangoDB batch job {}>'.format(self._id)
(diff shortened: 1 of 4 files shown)


dancollins34:
The coveralls issues are some problems in the Pragma section I didn't write, and the python 2 only code.  Im not sure how to make it ignore the python 2 only section, or treat it differently

arango/batch.py
```diff
@@ -102,12 +122,22 @@ def commit(self):
         for later retrieval via its :func:`arango.batch.BatchJob.result`
         method
 
+        :returns: list of arango.batch.BatchJob or None
+
```
joowani: Missing the return type in the docstring. Please add:
```python
:rtype: [arango.batch.BatchJob] | None
```

arango/batch.py
```diff
@@ -102,12 +122,22 @@ def commit(self):
...
         :raises arango.exceptions.BatchExecuteError: if the batch request
             cannot be executed
         """
+
+        res = self._lock.acquire(timeout=self._batch_submit_timeout)
```
joowani: Maybe we should rename `batch_submit_timeout` to `lock_acquire_timeout` for clarity.

arango/batch.py
```diff
@@ -102,12 +122,22 @@ def commit(self):
...
+
+        if not res:
+            raise BatchExecuteError("Error: Unable to reaccquire lock within time period."
```
joowani: No need to start with `Error:` as it is already an exception message. This formatting is not consistent with other error messages.

arango/batch.py
```diff
@@ -214,6 +247,32 @@ def graph(self, name):
         return Graph(self, name)
 
 
+if six.PY2:
+    class Python2Lock(object):
+        # https://stackoverflow.com/questions/35149889/lock-with-timeout-in-python2-7
+        def __init__(self):
+            self._underlying = Queue.Queue(maxsize=1)
```
joowani: I think `_underlying` should be changed to  `_queue` for just straightforward clarity.

arango/batch.py
```diff
@@ -214,6 +247,32 @@ def graph(self, name):
...
+            self._underlying.put(True, block=False)
+
+        def acquire(self, timeout=0):
+            if timeout > 0:
+                try:
+                    return self._underlying.get(block=True, timeout=timeout)
+                except Queue.Empty:
+                    return False
+            else:
```
joowani: It is a common practice avoid using `else`. I personally think it's more readable that way.

Also, the default value of `timeout` would be better as `None` since we never use the value 0. It is less misleading (makes it clear to the users that they must set it to put it to use), and stays more consistent with how the rest of the library treats default values.

So I would change this method like this:

```python
def acquire(self, timeout=None):
    if not timeout:
        return self._underlying.get(block=True)
    try:
        return self._underlying.get(block=True, timeout=timeout)
    except Queue.Empty:
        return False
```
Or this would be even better if it's correct:
```python
def acquire(self, timeout=None):
    try:
        return self._underlying.get(block=True, timeout=timeout)
    except Queue.Empty:
        return False
```

arango/batch.py
```diff
@@ -225,6 +284,8 @@ def __init__(self):
         self._status = 'pending'
         self._result = None
```
joowani: Please remove the extra newline

arango/batch.py
```diff
@@ -225,6 +284,8 @@ def __init__(self):
...
 
+        self._lock = Lock()
```
joowani: How come we are not doing the python2 vs 3 check here?

arango/database.py
```diff
@@ -426,7 +426,7 @@ def asynchronous(self, return_result=True):
         """
         return AsyncExecution(self._conn, return_result)
 
-    def batch(self, return_result=True, commit_on_error=True):
+    def batch(self, return_result=True, commit_on_error=True, batch_submit_timeout=10):
```
joowani: Again, `lock_acquire_timeout` sounds clearer.

arango/batch.py
```diff
@@ -1,7 +1,8 @@
 from __future__ import absolute_import, unicode_literals
 
-from uuid import uuid4
+import six
 
+from uuid import uuid4
```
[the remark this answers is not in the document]
dancollins34: Sure, will do in the future

arango/batch.py
```diff
@@ -10,6 +11,11 @@
 from arango.response import Response
 from arango.aql import AQL
 
+if six.PY2:
+    import Queue
+
+from threading import Lock
```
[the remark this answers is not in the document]
dancollins34: I'll do that tonight

arango/batch.py
```diff
@@ -28,13 +34,16 @@ class BatchExecution(Connection):
         so far are committed even if an exception is raised before existing
         out of the context (default: ``False``)
     :type commit_on_error: bool
+    :param batch_submit_timeout: the timeout to use for acquiring the lock necessary to submit a batch.  Only relevant
```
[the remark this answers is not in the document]
dancollins34: also fine.  And i will do that in the future

arango/batch.py
```diff
@@ -28,13 +34,16 @@ class BatchExecution(Connection):
...
+        in multi-threaded contexts- in single threaded contexts, acquiring this lock will never fail
```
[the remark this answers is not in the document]
dancollins34: yep

arango/batch.py
```diff
@@ -28,13 +34,16 @@ class BatchExecution(Connection):
...
+    :type batch_submit_timeout: int
 
     .. warning::
```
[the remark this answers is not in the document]
dancollins34: agreed

arango/batch.py
```diff
@@ -28,13 +34,16 @@ class BatchExecution(Connection):
...
         Batch execution is currently an experimental feature and is not
         thread-safe.
     """
 
-    def __init__(self, connection, return_result=True, commit_on_error=False):
+    def __init__(self, connection, return_result=True, commit_on_error=False, batch_submit_timeout=10):
```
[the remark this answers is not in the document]
dancollins34: Sure. I can make it a constant in the batch.py file as well

arango/batch.py
```diff
@@ -85,14 +101,18 @@ def handle_request(self, request, handler):
...
         """
-        self._requests.append(request)
-        self._handlers.append(handler)
 
-        if not self._return_result:
-            return None
-        batch_job = BatchJob()
-        self._batch_jobs.append(batch_job)
-        return batch_job
+        to_return = None
```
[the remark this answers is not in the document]
dancollins34: The case i was envisioning was an aggregator- where one thread owns the batch, tracks its size, and submits it every time it gets to a given size.  However, there is currently no way to get the output of a batch if you are that managing process, since the batch contents are deleted after commit.  It also wouldnt be a bad thing to not have to keep track of your commits if you were going to put each of the BatchJobs into a list anyway.

arango/batch.py
```diff
@@ -102,12 +122,22 @@ def commit(self):
         for later retrieval via its :func:`arango.batch.BatchJob.result`
         method
 
+        :returns: list of arango.batch.BatchJob or None
```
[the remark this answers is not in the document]
dancollins34: Will do

arango/batch.py
dancollins34: yep

arango/batch.py
dancollins34: But its not a timeout for the lock being acquired, its a timeout for the lock being acquired ONLY when submitting.  The end user doesn't care about the lock, they care that its the amount of time it would take until their batch either starts to submit or raises an error

arango/batch.py
dancollins34: Okay.

arango/batch.py
dancollins34: Sure

arango/batch.py
dancollins34: Although python has duck typing, I dislike using if statements to check "falsy" objects.  That being said, your second suggestion would be correct since the default value of timeout is None

arango/batch.py
dancollins34: Sure

arango/batch.py
dancollins34: It doesn't need a timeout.  Python 2 has Lock, it just doesnt have a timeout

arango/batch.py
joowani: Sure that makes sense.

arango/batch.py
joowani: Got it.

arango/database.py
joowani: nvm this.

dancollins34:
I updated to account for your edits, also ran flake8

dancollins34:
ah the conditional import is causing errors. one sec

dancollins34:
The lock interface for python 3 requires that timeouts <= 0 wait forever, and the Queue interface for python 2 requires that timeout = None waits forever, so lock has to have an if statement to make it consistent with the python 3 lock

dancollins34:
done committing on this I think

PR #71, opened 2018-02-01
Add cluster-only properties for collections

chrised:
And only include them in the output if they actually exist


chrised:
Commit d13e3842245df27c6f8ecfa0f79a45550f3c0cd3 appears to work from my local testing against both an unclustered and clustered instance, but it looks like coverage is dropped if I run this against an unclustered deployment, and many things appear to go horribly wrong when I try and run the test suite against a clustered deployment.

PR #84, opened 2018-10-11
Honor read and write params to begin_transaction

gmacon:
In my application, I saw an error where a transaction containing only AQL would fail because the write collections were not declared. I traced this to the `read` and `write` parameters to `begin_transaction` not being passed all the way through to the database in `commit`.


joowani:
Thank you for the PR @gmacon

PR #122, opened 2020-02-11
Remove duplicate '_count' slot from Cursor

153957:
The `'_count'` slot was already defined on line 35.


joowani:
Thanks for the PR @153957!

PR #132, opened 2020-06-16
Fixes view links format

Darkheir:
When retrieving a view property the `links` attribute is a dict and not a list.

The key is the collection name and the value the link attributes.


joowani:
Thanks for the PR @Darkheir!

PR #143, opened 2020-09-01
Allow configuring schema property for existing collections

153957:
Although not entirely clear from this documentation: https://www.arangodb.com/docs/3.7/data-modeling-collections-collection-methods.html#properties

This page indicates it should be possible to set the schema property for existing collections: https://www.arangodb.com/docs/3.7/data-modeling-documents-schema-validation.html


joowani:
Thanks for the PR @153957!

PR #157, opened 2021-02-25
Warning about simple queries

wshayes:
Just saw you added this - really nice listing - and useful when you can use them.

I started using them and then had to back out as they aren't dependable. I didn't realize how limited they were.

Your library and docs just keep getting better and better. Thank you!


joowani:
Hi @wshayes,

Thanks for the contribution! I'm glad you find this library useful.

PR #165, opened 2021-04-14
fix smart graph creation

hkernbach:
SmartGraphs needs a "smartGraphAttribute" to be defined. 
Additionally, this needs to be set inside the options object `{options: {smartGraphAttribute: <attribute>}}`.
Same for `numberOfShards`.

### DIFF
--- a/arango/database.py
+++ b/arango/database.py
@@ -1209,7 +1209,7 @@ class Database(ApiGroup):
                 'to_vertex_collections': ['lectures']
             }
         """
-        data: Json = {"name": name}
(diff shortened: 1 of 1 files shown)


hkernbach:
Another comment: `replicationFactor` and `writeConcern` also can be defined within the options. 
Will cover that one in another PR or create a ticket :)

hkernbach:
Another one: Also support for disjoint smart graphs can be added :)

joowani:
Thanks for the contribution @hkernbach!

PR #168, opened 2021-05-17
[Fix] Explaining a query with bindings requires bindVars to be set.

aquamatthias:
Allow to provide bind_vars as parameter for `explain` and pass it to the request.


aquamatthias:
@joowani any chance to have this fix in a release?

joowani:
Hi @aquamatthias,

Sorry for the delay. I released [7.2.0](https://github.com/joowani/python-arango/releases/tag/7.2.0) including the fix. Thanks.

PR #180, opened 2021-10-19
user defined REQUEST_TIMEOUT

alexbakharew:
Hello! I added parameter to HTTP client for request_timeout.

Currently i am working on benchmarks for comparing ArangoSearch and ElasticSearch. And sometimes i need to insert big batch of document to remote server. It takes more than 60s to do this. That is why I need to set this parameter by myself

### DIFF
--- a/arango/client.py
+++ b/arango/client.py
@@ -42,6 +42,8 @@ class ArangoClient:
         the de-serialized object. If not given, ``json.loads`` is used by
         default.
     :type deserializer: callable
+    :param request_timeout: User-defined time in seconds for performing 
+        http requsts. 60s by default
     """
 
     def __init__(
@@ -51,6 +53,7 @@ class ArangoClient:
         http_client: Optional[HTTPClient] = None,
         serializer: Callable[..., str] = lambda x: dumps(x),
         deserializer: Callable[[str], Any] = lambda x: loads(x),
+        request_timeout: Optional[int] = 60
     ) -> None:
         if isinstance(hosts, str):
             self._hosts = [host.strip("/") for host in hosts.split(",")]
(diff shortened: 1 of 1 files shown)


alexbakharew:
Thank you, @joowani, I created custom http client with extended timeout. And it works fine
But I think that it is a little bit complicate to define your own implementation of http client only for this small option. Also I need to implement following abstract methods: arango.http.HTTPClient.create_session() and arango.http.HTTPClient.send_request().

What do you think about it?

joowani:
Sorry for the late reply @alexbakharew, not sure why I didn't get any notifications for your replies.

I see where you are coming from. But once we add a field for `timeout`, what about other fields for configuring certs, authentication, `x-` headers etc? Should we add them too? And what if we wanted to swap out `requests` with a different HTTP library altogether? 

The abstract classes are there to allow the users switch to a different library if needed, and not tie down the ArangoClient signature with the `requests` library (which may or may not always be the default in the future).

joowani:
Hi again @alexbakharew,

After some more thought, I think that adding the timeout parameter to ArangoClient signature is ok, as long it is passed into the HTTP client in a more "generic way" (e.g. in a dict named `params`). Ultimately I'll leave it up to you and the ArangoDB team.

alexbakharew:
Hi @joowani, I also missed your previous reply.

Sound good! You will implement it by yourself?

joowani:
Hi @alexbakharew, 

I would like to eventually, but at the moment I'm in the middle of a big transition and will not be able to actively work on any features for at least another couple of months. I suppose your PR will be ok for the time being (my suggested approach should not change the API in the future).

joowani:
I'll revisit this issue another time. But for now let's stick with CustomHTTPClients.

PR #184, opened 2021-12-03
feature: host fallback support

### LINKED ISSUE #179: Support for Fallback hosts
There is an option to pass multiple hosts while creating AragoClient. But if we pass two hosts and one of the server is down, the call to connect fails with the below error 

For example, 

`client_hosts = ["http://127.0.0.1:8529", "http://localhost:9539"]
http_client = DefaultHTTPClient()
client2 = ArangoClient(
        hosts=client_hosts,
        host_resolver="random",
        http_client=http_client,
        serializer=json.dumps,
        deserializer=json.loads,
    )

db = client2.db('test', username='root', password='')
`

In the above snippet, the 2nd hosts to client_hosts http://localhost:9539 is an host that is down, the above throws below error 

`File "/usr/local/lib/python3.9/site-packages/requests/adapters.py", line 516, in send
    raise ConnectionError(e, request=request)
requests.exceptions.ConnectionError: HTTPConnectionPool(host='localhost', port=9529): Max retries exceeded with url: /_db/test/_api/collection (Caused by NewConnectionError('<u
rllib3.connection.HTTPConnection object at 0x1091a7b80>: Failed to establish a new connection: [Errno 61] Connection refused'))`

Is it possible to implement the fallback host like we have in Java driver - https://www.arangodb.com/docs/stable/drivers/java-reference-setup.html#fallback-hosts

aMahanna:
Closes #179 

A (working) first-draft implementation for host fallback support.

Requires feedback

What's new:
* `BaseConnection.process_request()`
    * Given a request, process it until a valid response is provided, or the max amount of tries has been reached.
    * `host_index` is still being chosen using the resolver's `get_host_index` method.
* `HostResolver.max_tries : int`
    * A limit to the amount of attempts made to execute the request. Value varies based on the resolver type
* `HostResolver.host_count : int`
    * The number of hosts provided
* `RandomHostResolver.get_host_index()`
    * Reworked to keep track of previously chosen host indexes, to make sure that the next chosen host_index is not one of them.

### DIFF
--- a/arango/client.py
+++ b/arango/client.py
@@ -66,7 +66,7 @@ class ArangoClient:
         self._host_resolver: HostResolver
 
         if host_count == 1:
-            self._host_resolver = SingleHostResolver()
+            self._host_resolver = SingleHostResolver(1, resolver_max_tries)
         elif host_resolver == "random":
             self._host_resolver = RandomHostResolver(host_count, resolver_max_tries)
         else:
--- a/arango/client.py
+++ b/arango/client.py
@@ -31,6 +31,10 @@ class ArangoClient:
         multiple host URLs are provided). Accepted values are "roundrobin" and
         "random". Any other value defaults to round robin.
     :type host_resolver: str
+    :param resolver_max_tries: Number of attempts to process an HTTP request
+        before throwing a ConnectionAbortedError. Must not be lower than the
+        number of hosts.
+    :type resolver_max_tries: int
     :param http_client: User-defined HTTP client.
     :type http_client: arango.http.HTTPClient
     :param serializer: User-defined JSON serializer. Must be a callable
@@ -48,6 +52,7 @@ class ArangoClient:
         self,
         hosts: Union[str, Sequence[str]] = "http://127.0.0.1:8529",
         host_resolver: str = "roundrobin",
+        resolver_max_tries: Optional[int] = None,
         http_client: Optional[HTTPClient] = None,
         serializer: Callable[..., str] = lambda x: dumps(x),
         deserializer: Callable[[str], Any] = lambda x: loads(x),
@@ -63,9 +68,9 @@ class ArangoClient:
         if host_count == 1:
             self._host_resolver = SingleHostResolver()
         elif host_resolver == "random":
-            self._host_resolver = RandomHostResolver(host_count)
+            self._host_resolver = RandomHostResolver(host_count, resolver_max_tries)
         else:
-            self._host_resolver = RoundRobinHostResolver(host_count)
+            self._host_resolver = RoundRobinHostResolver(host_count, resolver_max_tries)
 
         self._http = http_client or DefaultHTTPClient()
         self._serializer = serializer
--- a/arango/connection.py
+++ b/arango/connection.py
@@ -140,7 +140,7 @@ class BaseConnection:
                 return self.prep_response(resp, request.deserialize)
             except ConnectionError:
                 url = self._url_prefixes[host_index] + request.endpoint
-                logging.exception(f"ConnectionError: {url}")
+                logging.error(f"ConnectionError: {url}")
 
                 if len(indexes_to_filter) == self._host_resolver.host_count - 1:
                     indexes_to_filter.clear()
--- a/arango/connection.py
+++ b/arango/connection.py
@@ -9,10 +9,10 @@ __all__ = [
 import sys
 import time
 from abc import abstractmethod
-from typing import Any, Callable, Optional, Sequence, Union
+from typing import Any, Callable, Optional, Sequence, Tuple, Union
 
 import jwt
-from requests import Session
+from requests import ConnectionError, Session
 from requests_toolbelt import MultipartEncoder
 
 from arango.exceptions import JWTAuthError, ServerConnectionError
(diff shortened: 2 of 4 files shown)


arango/connection.py
```diff
@@ -110,6 +110,42 @@ def prep_response(self, resp: Response, deserialize: bool = True) -> Response:
         resp.is_success = http_ok and resp.error_code is None
         return resp
 
+    def process_response(
```
joowani, line 113: Something like `process_request` sounds more suitable here.

arango/resolver.py
```diff
@@ -7,20 +7,24 @@
...
 
 class HostResolver(ABC):  # pragma: no cover
     """Abstract base class for host resolvers."""
 
+    def __init__(self) -> None:
+        self._max_tries: int = 3
+
     @abstractmethod
-    def get_host_index(self) -> int:
+    def get_host_index(self, prev_host_index: Optional[int] = None) -> int:
```
joowani, line 20: To keep the `*Resolver` classes as stateless as possible, I recommend passing in a `Set[int]` here (e.g. "indexes_to_filter") instead of `prev_host_index`. And you can maintain the `Set[int]` in `process_response`.

arango/resolver.py
```diff
@@ -7,20 +7,24 @@
...
 import random
 from abc import ABC, abstractmethod
+from typing import Optional, Set
 
 
 class HostResolver(ABC):  # pragma: no cover
     """Abstract base class for host resolvers."""
 
+    def __init__(self) -> None:
+        self._max_tries: int = 3
```
joowani, line 17: Should this value be user-configurable?

arango/connection.py
```diff
@@ -110,6 +110,42 @@ def prep_response(self, resp: Response, deserialize: bool = True) -> Response:
...
+
+        :param host_index: The index of the first host to try
+        :type host_index: int
+        :param request: HTTP request.
+        :type request: arango.request.Request
+        :return: HTTP response.
+        :rtype: arango.response.Response
+        """
+        tries = 0
+        while tries < self._host_resolver._max_tries:
```
joowani, line 126: nit: accessing private variable `_max_tries`. You can use something like `@property` for `max_tries`.

arango/connection.py
```diff
@@ -110,6 +110,42 @@ def prep_response(self, resp: Response, deserialize: bool = True) -> Response:
...
+
+                return self.prep_response(resp, request.deserialize)
+            except ConnectionError:
+                tries += 1
+                host_index = self._host_resolver.get_host_index(
+                    prev_host_index=host_index
+                )
+
+        raise ConnectionAbortedError(
+            "Unable to establish connection to host(s) within limit"
```
joowani, line 146: Might be helpful to include the limit value in the error message.

arango/connection.py
```diff
@@ -110,6 +110,42 @@ def prep_response(self, resp: Response, deserialize: bool = True) -> Response:
...
+                    method=request.method,
+                    url=self._url_prefixes[host_index] + request.endpoint,
+                    params=request.params,
+                    data=self.normalize_data(request.data),
+                    headers=request.headers,
+                    auth=auth,
+                )
+
+                return self.prep_response(resp, request.deserialize)
+            except ConnectionError:
```
joowani, line 139: We are not preserving the contents of ConnectionErrors (e.g. error messages) here, which may hinder debugging. A few options I can think of are 1) calling `logger.exception` 2) pass down the contents of ConnectionErrors in ConnectionAbortedError (seems like the more difficult option).

joowani (commented):
Looks like a good start. I left some comments, some of which may not need to be addressed immediately if you are looking to do incremental changes. 

Also, I'm not sure how much of this overlaps with active failover in ArangoDB (which python-arango does not have support for yet). I started a [thread](https://github.com/arangodb/arangodb/issues/13860) on it in the past but never got a chance to work on it. Hopefully that helps you in the future.

arango/resolver.py
aMahanna: Makes sense. I've addressed what I think you're looking for, we maintain the `Set[int]` in `process_response`: https://github.com/ArangoDB-Community/python-arango/pull/184/files#diff-ff65f7eb9f247417f86b688f08c1dc65226bfe8298866e685349f3a902c062bdR144-R146

arango/resolver.py
aMahanna: Good call, [it is now](https://github.com/ArangoDB-Community/python-arango/pull/184/files#diff-d72fa74f337fd1eb022878d3e3f0f8889bfa8872a933a67e851f37c9160d535dR34-R37)

arango/connection.py
aMahanna: Right, I'm not 100% sure what you're looking for, but I've added [this line](https://github.com/ArangoDB-Community/python-arango/pull/184/files#diff-ff65f7eb9f247417f86b688f08c1dc65226bfe8298866e685349f3a902c062bdR142) in our `except` clause. Can you confirm that this is what you were thinking of having?

arango/client.py
```diff
@@ -31,6 +31,10 @@ class ArangoClient:
         multiple host URLs are provided). Accepted values are "roundrobin" and
         "random". Any other value defaults to round robin.
     :type host_resolver: str
+    :param resolver_max_tries: Number of attempts to process an HTTP request
+        before throwing a ConnectionAbortedError. Must not be lower than the
+        number of hosts.
```
joowani, line 36: We should add a check for number of hosts < max tries.

arango/resolver.py
```diff
@@ -13,51 +13,54 @@
...
+    def __init__(self, host_count: int = 1, max_tries: int = 3) -> None:
+        if max_tries < host_count:
+            raise ValueError("max_tries cannot be less than host_count")
+
+        self._host_count = host_count
+        self._max_tries = max_tries
 
     @abstractmethod
-    def get_host_index(self, prev_host_index: Optional[int] = None) -> int:
+    def get_host_index(self, indexes_to_filter: Set[int] = set()) -> int:
```
joowani, line 25: Using a mutable object like a set as default arguments should be avoided ([reference](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments)). I recommend using `None` here instead and initialize the set in the function.

arango/connection.py
joowani: Yes that's correct.

arango/client.py
aMahanna: currently we have this over [here](https://github.com/ArangoDB-Community/python-arango/pull/184/files#diff-f698f829422bf3c3602d6df4a9b8d137054a74c366a6d700aedcf8854c92efc0R18-R19), but I can move it to `client.py` if you'd like?

aMahanna: After giving it some thought I think it would be best if we leave it within `resolver.py`

aMahanna: After giving it some thought I think it would be best if we left it within `resolver.py`

arango/connection.py
aMahanna: I'll switch this to `logging.error` for now to avoid the noise generated by `logging.exception`

arango/client.py
joowani: Oh right. Yes I think it's ok to leave it in `resolver.py`.

arango/connection.py
joowani: Good point. But did you mean to say `logging.debug`? Usually we use lower level messages to allow users to reduce noise.

aMahanna: Ah good call, I've updated that

arango/client.py
aMahanna: sounds good!

joowani (approved):
Lgtm. Thanks for your contribution!

PR #189, opened 2021-12-23
new: fallback host support for 503 errorNum

aMahanna:
From a discussion with the ArangoDB Java & Go maintainers (cc @rashtao):

> please note that failover/retry in case of status 503 should happen only in these cases:
     1. the response has body like {"error":true,"errorNum":503,"code":503, ...}, note errorNum=503
     2. the 503 response comes from an intermediate proxy, so it does not contain a json body like the one above
Other 503 responses  with different errorNum should not lead to failover/retry, but just thrown to the caller.

### DIFF
--- a/arango/connection.py
+++ b/arango/connection.py
@@ -104,6 +104,8 @@ class BaseConnection:
             if isinstance(resp.body, dict):
                 resp.error_code = resp.body.get("errorNum")
                 resp.error_message = resp.body.get("errorMessage")
+                if resp.status_code == resp.error_code == 503:
+                    raise ConnectionError  # Fallback to another host
         else:
             resp.body = resp.raw_body
 


arango/connection.py
```diff
@@ -104,6 +104,8 @@ def prep_response(self, resp: Response, deserialize: bool = True) -> Response:
             if isinstance(resp.body, dict):
                 resp.error_code = resp.body.get("errorNum")
                 resp.error_message = resp.body.get("errorMessage")
+                if resp.status_code == resp.error_code == 503:
+                    raise ConnectionError  # Fallback to another host
         else:
             resp.body = resp.raw_body
```
aMahanna, line 110: Just curious, in what situations would we not want to deserialize the response raw body?

joowani: Good question. Although I can't remember exactly which HTTP API endpoints, some did not return JSON objects but strings instead.

arango/connection.py
```diff
@@ -104,6 +104,8 @@ def prep_response(self, resp: Response, deserialize: bool = True) -> Response:
             if isinstance(resp.body, dict):
                 resp.error_code = resp.body.get("errorNum")
                 resp.error_message = resp.body.get("errorMessage")
+                if resp.status_code == resp.error_code == 503:
+                    raise ConnectionError  # Fallback to another host
```
joowani, line 108: I think `process_request` is the better place for this logic (at line 141). Also you do not need to check `error_code`, as it is auto-populated from `status_code` for HTTP 503. 

`prep_response` is only for sprinkling few convenience goodies into the response object.

joowani (changes requested):

arango/connection.py
aMahanna: > Also you do not need to check error_code, as it is auto-populated from status_code for HTTP 503

Sorry, I'm not sure what you mean here

For example, let's say we have the following HTTP 503 response:
```json
{
   "code":503,
   "error":true,
   "errorNum":1478,
   "errorMessage":"AQL: A cluster backend which was required for the operation could not be reached. Please check: PRMR-7267a1da-0182-4465-b5fd-53043b6f26c3 (while optimizing plan)"
}
```

We would need access to both the `code` and `errorNum` attributes to check whether this HTTP 503 falls into the umbrella of default fallback host support. In this response case, we would throw the error to the user (and not rely on fallback host support), since `errorNum != 503`

We only want to provide fallback host support if we get the following HTTP 503 response:
```json
{
   "code":503,
   "error":true,
   "errorNum":503,
   "errorMessage":"(...)"
}
```

In order for us to check this edge case, we would need access to the deserialized JSON body, to fetch the `errorNum` attribute (in which case you already do in `prep_response()` (line 105).

Let me know if I'm missing something, I would also prefer if this was in `process_request()` but I believe I would have to add in some extra deserialize logic, which could become redundant, given the behaviour of `prep_response()`.

arango/connection.py
aMahanna: Ah good to know, thank you 👍

arango/connection.py
joowani: As far as I know, the `code` field in the JSON body is just a "replica" of the HTTP status code. The `errorNum` field can also be a "replica", or sometimes it is an internal error number specific to ArangoDB (ArangoDB team should be more knowledgeable here). You can access the HTTP status code directly from attribute `Response.status_code` and do not need to deserialize or use the body at all.

You can take a look at [here](https://github.com/ArangoDB-Community/python-arango/blob/main/arango/http.py) to see how the Response object is created.

aMahanna: I've double checked with @rashtao (Java driver maintainer): 

> The json body field code should be the same as the http response code, and the json body errorNum is an ArangoDB internal code. So I think your code is fine as is.

We _only_ want Fallback Host to cover an HTTP 503 where its `errorNum` also happens to be 503 (not always the case, i.e status_code is different from error_code, and we need both). To do that, we would need access to the deserialized body.

If we only check that the status code of the response is 503, then we would be adding Fallback Host support to _all_ errorNums associated to an HTTP 503, which is not consistent with what the Java & Go drivers are doing by default.

aMahanna: (The value of `error_code` is the `errorNum` attribute from the JSON body: https://github.com/ArangoDB-Community/python-arango/blob/281797b827e1ff2cd51254af7a89c83be5c34225/arango/connection.py#L105)

joowani: Ah I see. Thanks for clarifying I misread one of your previous comments. I still don't like how it is in prep_response but I this is fine for now.

joowani (approved):

PR #192, opened 2022-01-21
feat: Use HEAD HTTP method to check if a document exists

Darkheir:
The current `has` method requests the full document to only check if it exists or not. It may be overkill for big documents.

This PR replaces it with a simple `HEAD` call allowing to know if the document exists or not.

:warning:  This PR will change the `error_code` returned. Since we don't have a body in the response, the `error_code` value will default to `status_code`.
If the only codes we should handle are the ones in the test, I could update the response object to set the appropriate `error_code` based on the status code.

### DIFF
--- a/arango/collection.py
+++ b/arango/collection.py
@@ -637,8 +637,6 @@ class Collection(ApiGroup):
         )
 
         def response_handler(resp: Response) -> bool:
-            if resp.error_code == 1202:
-                return False
             if resp.status_code == 412:
                 raise DocumentRevisionError(resp, request)
             if resp.status_code == 404:
--- a/arango/collection.py
+++ b/arango/collection.py
@@ -630,7 +630,7 @@ class Collection(ApiGroup):
             headers["x-arango-allow-dirty-read"] = "true"
 
         request = Request(
-            method="get",
+            method="head",
             endpoint=f"/_api/document/{handle}",
             headers=headers,
             read=self.name,
@@ -641,9 +641,11 @@ class Collection(ApiGroup):
                 return False
             if resp.status_code == 412:
                 raise DocumentRevisionError(resp, request)
+            if resp.status_code == 404:
+                return False
             if not resp.is_success:
                 raise DocumentInError(resp, request)
-            return bool(resp.body)
+            return True
 
         return self._execute(request, response_handler)
 
--- a/tests/test_document.py
+++ b/tests/test_document.py
@@ -1554,7 +1554,7 @@ def test_document_has(col, bad_col, docs):
 
         with assert_raises(DocumentRevisionError) as err:
             col.has(doc_input, rev=bad_rev, check_rev=True)
-        assert err.value.error_code in {412, 1200}
+        assert err.value.error_code == 412
 
     # Test existing documents with bad revision
     for doc_input in [
@@ -1564,15 +1564,15 @@ def test_document_has(col, bad_col, docs):
     ]:
         with assert_raises(DocumentRevisionError) as err:
             col.has(doc_input)
-        assert err.value.error_code in {412, 1200}
+        assert err.value.error_code == 412
 
         with assert_raises(DocumentRevisionError) as err:
             col.has(doc_input, rev=bad_rev)
-        assert err.value.error_code in {412, 1200}
+        assert err.value.error_code == 412
 
         with assert_raises(DocumentRevisionError) as err:
             col.has(doc_input, rev=bad_rev, check_rev=True)
-        assert err.value.error_code in {412, 1200}
+        assert err.value.error_code == 412
 
         assert doc_input in col
         assert col.has(doc_input, rev=rev, check_rev=True) is True
@@ -1651,12 +1651,12 @@ def test_document_has(col, bad_col, docs):
     # Test get with bad database
     with assert_raises(DocumentInError) as err:
         bad_col.has(doc_key)
-    assert err.value.error_code in {11, 401}
+    assert err.value.error_code == 401
 
     # Test contains with bad database
     with assert_raises(DocumentInError) as err:
         assert doc_key in bad_col
-    assert err.value.error_code in {11, 401}
+    assert err.value.error_code == 401
 
 
 def test_document_get(col, bad_col, docs):
--- a/tests/test_document.py
+++ b/tests/test_document.py
@@ -1554,7 +1554,7 @@ def test_document_has(col, bad_col, docs):
 
         with assert_raises(DocumentRevisionError) as err:
             col.has(doc_input, rev=bad_rev, check_rev=True)
-        assert err.value.error_code == 1200
+        assert err.value.error_code in {412, 1200}
 
     # Test existing documents with bad revision
     for doc_input in [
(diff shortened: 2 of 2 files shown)


joowani:
Hi @Darkheir, thanks for the PR. Could you please address the failing test? Thanks.

Darkheir:
After some investigation it seems that ArangoDB server hangs on result retrieval for HEAD requests. I created an issue here: https://github.com/arangodb/arangodb/issues/15639

To solve this issue I created a specific condition in the AsyncExecutor for HEAD requests where we directly get the result and return a dummy async job that imitate the async job signature.

IMO this is not a big issue since HEAD requests are never going to be costly and the only one I know is for the document's headers.

joowani:
Hi @Darkheir,

I'm not in favor of introducing the dummy async job class. It seems too "hacky" for a small performance gain, and breaks the users' expectation of error codes from async API calls. I recommend that we undo the dummy class addition, and merge just the "head" bit once [arangodb/arangodb#15639](https://github.com/arangodb/arangodb/issues/15639) is resolved. Thanks again for the pull request and for creating the issue with ArangoDB.

Darkheir:
> Hi @Darkheir,
> 
> I'm not in favor of introducing the dummy async job class. It seems too "hacky" for a small performance gain, and breaks the users' expectation of error codes from async API calls. I recommend that we undo the dummy class addition, and merge just the "head" bit once [arangodb/arangodb#15639](https://github.com/arangodb/arangodb/issues/15639) is resolved. Thanks again for the pull request and for creating the issue with ArangoDB.

Fair enough, let's wait until ArangoDB fixed the issue :-) 

The bad thing is that even if ArangoDB fixes it, all the people using an old ArangoDB version will get some errors when using `python-arango`...

joowani:
That's a good point. You can leave this PR as is for now, and I'll think about it a bit more (maybe we'll merge this after all).

apetenchea:
Hi @Darkheir,

Thanks again for your work. Are you still interested in continuing this PR. Otherwise, I can take over.

The task is even easier now, as `DummyAsyncJob` is no longer needed.

Darkheir:
> Hi @Darkheir,
> 
> Thanks again for your work. Are you still interested in continuing this PR? Otherwise, I can take over.
> 
> The task is even easier now, as `DummyAsyncJob` is no longer needed.

Did ArangoDB fix the issue ? The ticket is still open on their side: #15639

apetenchea:
> Did ArangoDB fix the issue ? The ticket is still open on their side: #15639

3.8 is not supported since long ago. I tried the head method on 3.11 and it works fine. Anyway, the driver has a CI now, which will tell whether there's any problem, though I doubt there will be any. Since we're planning the next release to be a major one, this would be the time to make such a change.

Please let me know if you still want to take this through (no pressure).

Darkheir:
Just updated the PR, let's see how the CI goes :-)

arango/collection.py
```diff
@@ -640,9 +640,11 @@ def response_handler(resp: Response) -> bool:
                 return False
```
apetenchea, line 640: This will be no longer needed

tests/test_document.py
```diff
@@ -1553,7 +1553,7 @@ def test_document_has(col, bad_col, docs):
 
         with assert_raises(DocumentRevisionError) as err:
             col.has(doc_input, rev=bad_rev, check_rev=True)
-        assert err.value.error_code == 1200
+        assert err.value.error_code in {412, 1200}
```
apetenchea, line 1556: 1200 is no longer needed

apetenchea (commented):
Thanks! Change looks good and CI passed.

Because the response body for the "HEAD" method is always empty, as in this case "has" relies solely on return codes, could you please remove the special handing of the 1202 error code and 1200 from the tests? Just so we don't have obsolete code "rotting" in there.

Darkheir:
It seems that the CI doesn't run all the tests (no `--complete` flag).

The complete test suite is failing locally and I still reproduce #15639 on the latest version of ArangoDB...

apetenchea:
> It seems that the CI doesn't run all the tests (no `--complete` flag).
> 
> The complete test suite is failing locally and I still reproduce #15639 on the latest version of ArangoDB...

Ok, I'll take a look. Could you please let me know the pytest command you used? (to make sure we're talking about the same thing)

apetenchea:
My bad, I ran it without the `x-arango-async`. Indeed, the issue is still there 🙄 

I will look into the ArangoDB C++ code when I get some time, perhaps I can fix this issue. But I **really** want this to get through one way or another (worst case we can add an "use_get" parameter to make it work).

I cannot make any promise regarding the timing, but I am keeping this on my list and will get back to you at some point 🙏

apetenchea:
Hello @Darkheir,

I fixed the issue from the C++ codebase: https://github.com/arangodb/arangodb/pull/21087

This `HEAD` approach should work fine, starting from the next ArangoDB minor release. I would like to integrate your PR in the next driver release. I am aware that the next major driver release may still create issues for people using ArangoDB 3.11, in case anyone tries to send an async `HEAD` request, but that would be a very rare occurrence, to which there is a solution (keeping the driver on version 8, or using `get` instead of `has`).

If you're still interested in contributing (no pressure whatsoever), could you please:
- Update your PR so it gets all the changes from upstream.
- There is no longer an `error_code` associated with this type of `HEAD` request (because there is no body), but we'll rather work with `status_code`. That could be either `200`, `412` or `404`. Clean up the code handling of `1202`, as it is no longer necessary.
- Let's get rid of the "in" expressions from the tests (eg `assert err.value.error_code in {412, 1200}`) and instead use `==`. I think it should work just fine, and it looks much simpler.

Again, thanks for your initiative and the laaaarge amount of patience that was needed for this PR 😄

Darkheir:
Hi @apetenchea,

Is your fix included in a released version ? So I can make sure what I did is working fine.

Also it means that we would need to tests only for the latest version of ArangoDB and not the previous ones (right now both 3.11 and 3.12 are tested).


A less intrusive fix could be to introduce an intermediate release where depending on the server version a `HEAD` request is made or a  `GET` one. If we send a `GET` request we could also raise a warning saying that this will be soon removed and the server should be updated.

apetenchea:
> Is your fix included in a released version ? So I can make sure what I did is working fine.

Not yet, will be included in 3.12.1 once it gets released officially.

> Also it means that we would need to tests only for the latest version of ArangoDB and not the previous ones (right now both 3.11 and 3.12 are tested).

You just need to make sure the current CI passes. I'll take care of the async part, and will try it against the latest version compiled on my machine. But I know it works, I have tried something very similar to your PR already. We don't run the `--complete` tests here. Thanks for being proactive, but don't worry about that.

> A less intrusive fix could be to introduce an intermediate release where depending on the server version a `HEAD` request is made or a `GET` one. If we send a `GET` request we could also raise a warning saying that this will be soon removed and the server should be updated.

I don't think it's worth all the trouble. I like to believe that nobody is sending async HEAD requests, as that wouldn't bring anything but an overly complicated way of looking up documents. The async API is mainly intended for long running tasks (a complex query or a bulk insert), and using it for such a basic lookup would be like using a crane to go grocery shopping. Don't get me wrong, I'm all for backwards compatibility, but in this very specific case, a lot more users would benefit from sending a HEAD request than a GET, whether they're using 3.11 or 3.12. Those who really get into trouble just need to replace `has()` by `get()`. On the other hand, if you insist, I don't see any problem with your suggestion.

apetenchea:
Looks good to me! Thank you for you contribution!

I'll merge it tomorrow. It's finally going to happen 😄

apetenchea (approved):

PR #202, opened 2022-05-05
[Fix] Make overriding TLS verification optional

### LINKED ISSUE #201: Breaking change in TLS Handling in 7.3.3
After upgrading python-arango from 7.3.1 to 7.3.3 we are no longer able to connect to arangodb servers with a TLS setup using a custom CA certificate.
The related change: #199 overrides the verify flag of the HTTP session with a boolean value and does not respect what has been defined before.

We followed the instructions on the docs page to set up our own HttpClient: https://docs.python-arango.com/en/main/http.html
and provide the CA bundle to the session via the verify property. See code [ArangoDBClient](https://github.com/someengineering/resoto/blob/main/resotocore/resotocore/db/arangodb_extensions.py#L13-L44)

In order to fix the behavior I would suggest one of the following:
- remove the `verify_certificate` argument, since the same goal can be accomplished using the HttpClient
- make the `verify_certificate` argument optional and default to `None`

I am happy to provide a fix if you agree with one of the suggestions.

aquamatthias:
This PR allows overriding the TLS verification behavior.
It defaults to None and thus will have no effect if not explicitly defined.
If it is defined, it allows for enabling/disabling verification and defines a custom CA bundle file or directory as specified by the [requests](https://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification) library.

Fixes #201

### DIFF
--- a/arango/client.py
+++ b/arango/client.py
@@ -45,8 +45,13 @@ class ArangoClient:
         the de-serialized object. If not given, ``json.loads`` is used by
         default.
     :type deserializer: callable
-    :param verify_certificate: Verify TLS certificates.
-    :type verify_certificate: bool
+    :param verify_override: Override TLS certificate verification. This will
+       override the verify method of the underlying HTTP client.
+       None: Do not change the verification behavior of the underlying HTTP client.
+       True: Verify TLS certificate using the system CA certificates.
+       False: Do not verify TLS certificate.
+       str: Path to a custom CA bundle file or directory.
+    :type verify_override: Union[bool, str, None]
     """
 
     def __init__(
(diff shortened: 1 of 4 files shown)


aquamatthias:
@aMahanna added tests for this behavior. The PR that introduced the argument was already breaking.
My intuition for the parameter name: keeping `verify_certificate` makes it hard to understand the impact.
But I can also keep the name if you think it is better.

aMahanna:
> @aMahanna added tests for this behavior. The PR that introduced the argument was already breaking. My intuition for the parameter name: keeping `verify_certificate` makes it hard to understand the impact. But I can also keep the name if you think it is better.

Makes sense, `verify_override` is a more suitable name. We will move forward with a `8.0.0` release, and thank you for both the fix to https://github.com/ArangoDB-Community/python-arango/issues/201 and the added test case 👍

aMahanna (approved):

cw00dw0rd:
Thank you!

PR #206, opened 2022-06-13
Re-introduce setuptools-scm with correct usage

aMahanna:
### What's new
* Re-introduces the auto-versioning mechanic provided by `setuptools-scm` via `pyproject.toml` (instead of `setup.py`)

Some notes:
* Usage of `setuptools-scm` via `setup.py` has been [deprecated](https://github.com/pypa/setuptools_scm#setuppy-usage-deprecated) in favour of `pyproject.toml`
* `arango/version.py` has been deleted & reintroduced to the `.gitignore` file


arango/client.py
```diff
@@ -1,5 +1,12 @@
 __all__ = ["ArangoClient"]
 
+import sys
+
+if sys.version_info < (3, 8):
+    from importlib_metadata import version
+else:
+    from importlib.metadata import version
```
aMahanna, line 8: Unfortunately, importlib-metadata's import statement is not consistent, so this is required

arango/client.py
```diff
@@ -113,7 +119,7 @@ def version(self) -> str:
         :return: Client version.
         :rtype: str
         """
-        return version
+        return version("python-arango")  # type: ignore
```
aMahanna, line 122: A `# type: ignore` was added due to different `mypy` errors thrown for Python 3.7 / Python 3.8 / Python 3.9 builds.

Unfortunately, a `# type: ignore` [cannot be restricted to one specific python version](https://github.com/python/mypy/issues/8823). This results in having to set `warn_unused_ignores` in `pyproject.toml`. 

Will keep an eye out on the mypy issue.

aMahanna:
Two things that are annoyingly required for this re-introduction of `setuptools-scm` to work:
1. Setting `warn_unused_ignores` to false in `pyproject.toml`
2. Conditional import statement for `importlib-metadata` based on python version

aMahanna:
Currently trying to replace `importlib-metadata` with `pkg_resources`, but encountering `mypy`-related issues locally

arango/client.py
aMahanna: no longer relevant as of https://github.com/ArangoDB-Community/python-arango/pull/206/commits/4b3d067e97f715f9afa463622842bfc1ce593716

arango/client.py
aMahanna: No longer relevant as of https://github.com/ArangoDB-Community/python-arango/pull/206/commits/4b3d067e97f715f9afa463622842bfc1ce593716

joowani (approved):
Thanks for working on this.

PR #207, opened 2022-06-14
Add batch_size parameter in import_bulk method

aMahanna:
Introduces batch insertion via the `import_bulk` API.

These changes warrant a minor version increment.

### Demo

```py
vertices = [{'foo': str(i)} for i in range(1, 101)]
db.create_collection('numbers')
results = db.collection('numbers').import_bulk(vertices, batch_size=len(vertices)//10)
assert len(results) == 10
```

### DIFF
--- a/arango/collection.py
+++ b/arango/collection.py
@@ -2027,7 +2027,7 @@ class Collection(ApiGroup):
                 return result
             raise DocumentInsertError(resp, request)
 
-        result = []
+        results = []
         for batch in get_batches(documents, batch_size):
             request = Request(
                 method="post",
@@ -2037,9 +2037,9 @@ class Collection(ApiGroup):
                 write=self.name,
             )
 
-            result.append(self._execute(request, response_handler))
+            results.append(self._execute(request, response_handler))
 
-        return result[0] if len(result) == 1 else result
+        return results[0] if len(results) == 1 else results
 
 
 class StandardCollection(Collection):
--- a/arango/collection.py
+++ b/arango/collection.py
@@ -1985,10 +1985,15 @@ class Collection(ApiGroup):
         :type on_duplicate: str
         :param sync: Block until operation is synchronized to disk.
         :type sync: bool | None
-        :param batch_size: Max number of documents to import at once. If
-            unspecified, will import all documents at once. Note that the
-            output type changes to list[dict] if **batch_size** is specified.
-        :type batch_size: int | None
+        :param batch_size: Split up **documents** into batches of max length
+            **batch_size** and import them in a loop on the client side. If
+            **batch_size** is specified, the return type of this method
+            changes from a result object to a list of result objects.
+            IMPORTANT NOTE: this parameter may go through breaking changes
+            in the future where the return type may not be a list of result
+            objects anymore. Use it at your own risk, and avoid
+            depending on the return value if possible.
+        :type batch_size: int
         :return: Result of the bulk import.
         :rtype: dict | list[dict]
         :raise arango.exceptions.DocumentInsertError: If import fails.
--- a/arango/collection.py
+++ b/arango/collection.py
@@ -189,7 +189,7 @@ class Collection(ApiGroup):
             body = body.copy()
             body["_key"] = doc_id[len(self._id_prefix) :]
             return body
-        elif index:
+        elif index is not None:
             body = body.copy()
             body["_key"] = str(index)
             return body
--- a/arango/collection.py
+++ b/arango/collection.py
@@ -2022,19 +2022,29 @@ class Collection(ApiGroup):
                 return result
             raise DocumentInsertError(resp, request)
 
-        results = []
-        for batch in get_batches(documents, batch_size or len(documents)):
+        if batch_size is None:
             request = Request(
                 method="post",
                 endpoint="/_api/import",
-                data=batch,
+                data=documents,
                 params=params,
                 write=self.name,
             )
 
-            results.append(self._execute(request, response_handler))
+            return self._execute(request, response_handler)
+        else:
+            results = []
+            for batch in get_batches(documents, batch_size):
+                request = Request(
+                    method="post",
+                    endpoint="/_api/import",
+                    data=batch,
+                    params=params,
+                    write=self.name,
+                )
+                results.append(self._execute(request, response_handler))
 
-        return results[0] if batch_size is None else results
+            return results
 
 
 class StandardCollection(Collection):
--- a/arango/collection.py
+++ b/arango/collection.py
@@ -1986,10 +1986,11 @@ class Collection(ApiGroup):
         :param sync: Block until operation is synchronized to disk.
         :type sync: bool | None
         :param batch_size: Max number of documents to import at once. If
-            unspecified, will import all documents at once.
+            unspecified, will import all documents at once. Note that the
+            output type changes to list[dict] if **batch_size** is specified.
         :type batch_size: int | None
         :return: Result of the bulk import.
-        :rtype: dict
+        :rtype: dict | list[dict]
         :raise arango.exceptions.DocumentInsertError: If import fails.
         """
         documents = [self._ensure_key_from_id(doc) for doc in documents]
--- a/arango/collection.py
+++ b/arango/collection.py
@@ -169,18 +169,13 @@ class Collection(ApiGroup):
             else:
                 return doc_id, doc_id, {"If-Match": rev}
 
-    def _ensure_key_in_body(self, body: Json, index: Optional[int] = None) -> Json:
-        """Return the document body with "_key" field populated. If
-            neither "_key" or "_id" exist, set "_key" value to **index**,
-            where **index** is the document's position in the sequence.
-
+    def _ensure_key_in_body(self, body: Json) -> Json:
+        """Return the document body with "_key" field populated.
         :param body: Document body.
         :type body: dict
-        :param index: Document index value in the original list of documents.
-        :param index: int | None
         :return: Document body with "_key" field.
         :rtype: dict
-        :raise arango.exceptions.DocumentParseError: On missing _key, _id, & index.
+        :raise arango.exceptions.DocumentParseError: On missing ID and key.
         """
         if "_key" in body:
             return body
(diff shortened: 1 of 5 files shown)


aMahanna:
Looking for feedback on the data structure used for the `collections` parameter of `create_graph()`:

https://github.com/ArangoDB-Community/python-arango/blob/247ed1c8c7defb28054edf90616c9c5625d2496e/arango/database.py#L1240-L1264

aMahanna:
Q: Does `python-arango`'s readthedocs generate documentation based on anything else but the docstrings? Wondering if I should be aware of any other API documentation that I should be editing accordingly

arango/utils.py
```diff
@@ -82,3 +82,27 @@ def is_none_or_str(obj: Any) -> bool:
...
+    """
+    if batch_size is None or batch_size <= 0 or batch_size >= len(l):
+        return [l]
+
+    def generator() -> Iterator[Sequence[Json]]:
+        n = int(batch_size)  # type: ignore # (false positive)
+        for i in range(0, len(l), n):
+            yield l[i : i + n]
+
+    return generator()
```
joowani, line 108: I think it's reasonable to assume that when users are calling this function, they know the `batch_size` in advance (so we don't have to make it optional). And the code can be simplified to something like this:
```
def get_batches(elements: Sequence[Json], batch_size: int) -> Iterator[Sequence[Json]]]:
    for index in range(0, len(elements), batch_size):
        yield elements[index: index + batch_size]
```

arango/database.py
```diff
@@ -1217,18 +1218,49 @@ def create_graph(
...
+                    'docs': lecture_vertices_to_insert
+                    'options': {
+                        'overwrite' = False,
+                        'sync' = False,
+                        'batch_size' = 4
+                    }
+                },
+                'teach': {
+                    'docs': teach_edges_to_insert
+                }
```
joowani, line 1263: On a closer look, I am not sure if the `collections` parameter makes things easier than just calling the `impulk_bulk` methods on vertex and edge collections separately. In fact, I think it's less convenient because you won't be able to take advantage of argument docstring or type hints from IDEs while composing the dict argument.

arango/collection.py
```diff
@@ -1934,7 +1934,8 @@ def import_bulk(
         overwrite: Optional[bool] = None,
         on_duplicate: Optional[str] = None,
         sync: Optional[bool] = None,
-    ) -> Result[Json]:
+        batch_size: Optional[int] = None,
+    ) -> Union[Result[Json], List[Result[Json]]]:
```
joowani, line 1938: I'm inclined to add a new method "import_bulk _by_batches" just to avoid the `Union` type. I'm not a fan of a function returning different types if it can be avoided.

joowani (changes requested):

arango/database.py
aMahanna: cc @rajivsam

arango/utils.py
aMahanna: makes sense, assuming we move to `import_bulk_by_batches`

arango/collection.py
aMahanna: As an alternative to this, would it be worth looking into converting `List[Result[Json]]` into a `Result[Json]` (i.e combined all "batch" `import_bulk` results into one)

arango/utils.py
rajivsam: I agree with @joowani . Does it make sense to flag an error if this is not provided?

arango/database.py
rajivsam: The import bulk methods do not generate the id's for you I believe, so as long as we have a way to automatically create the ID's, that works for me.

arango/collection.py
joowani: Actually on second thought I think the way you have it is the best. Let's keep it but clearly document in the docstring that the output type will change when you set `batch_size`.

arango/database.py
joowani: Is there a strong reason why ID generation cannot live outside of python-arango (e.g. in your application code or in some shared utils package)?

Bulk generating keys from python-arango's side (without proper support from ArangoDB server for atomic operations) doesn't seem like a good idea due to key collision on concurrent requests or retry after failure, unless it's something like uuid4.

Supporting incremental and deterministic key generation for example, means python-arango would have to keep track of ID(s) internally. I think python-arango should aim to be as stateless as possible.

One solution here would be to allow users to pass in a `KeyGenerator` object on collection creation with method like `MyCustomKeyGenerator.generate_next_key()`. This way we would be placing the burden of about managing state and concurrency on the user.

arango/utils.py
joowani: > makes sense, but I assume this change is required only if we move to `import_bulk_by_batches`? Otherwise we would need to retain the current state of `get_batches` to support `batch_size=None`

Not sure if I understood, but we would be leaving `import_bulk` as is, and `batch_size` would be a required argument in `import_bulk_by_batches`.

arango/database.py
rajivsam: I like the key generator idea, works for me!

arango/collection.py
aMahanna: > Actually on second thought I think the way you have it is the best. Let's keep it but clearly document in the docstring that the output type will change when you set batch_size.

Addressed in 56d129f. Feel free to resolve if this change is suitable

arango/collection.py
```diff
@@ -1984,8 +1985,12 @@ def import_bulk(
         :type on_duplicate: str
         :param sync: Block until operation is synchronized to disk.
         :type sync: bool | None
+        :param batch_size: Max number of documents to import at once. If
+            unspecified, will import all documents at once. Note that the
+            output type changes to list[dict] if **batch_size** is specified.
+        :type batch_size: int | None
```
joowani, line 1991: If ArangoDB supports `batch_size` natively server-side one day, the return type may not be a `list[Result]` anymore (which would be a breaking change). Let's update the docstring as follows to warn the users:
```
Split up **documents** into batches of max length **batch_size** and import them in a loop on the client side. If **batch_size** is specified, the return type of this method changes from a result object to a list of result objects. IMPORTANT NOTE: this parameter may go through breaking changes in the future where the return type may not be a list of result objects anymore. Use it at your own risk, and avoid depending on the return value if possible. 
```

arango/utils.py
```diff
@@ -82,3 +82,16 @@ def is_none_or_str(obj: Any) -> bool:
...
     return obj is None or isinstance(obj, str)
+
+
+def get_batches(elements: Sequence[Json], batch_size: int) -> Iterator[Sequence[Json]]:
+    """Generator to split a list in batches
+        of (maximum) **batch_size** elements each.
+
+    :param elements: The list of elements.
+    :type elements: Sequence[Json]
+    :param batch_size: Number of elements per batch.
```
joowani, line 93: Let's change this to `Max number of elements per batch`.

arango/collection.py
```diff
@@ -2006,21 +2011,25 @@ def import_bulk(
...
+                method="post",
+                endpoint="/_api/import",
+                data=batch,
+                params=params,
+                write=self.name,
+            )
+
+            results.append(self._execute(request, response_handler))
+
+        return results[0] if len(results) == 1 else results
```
joowani, line 2032: There is a "bug" here. See my comment below in the unit tests.

Let's just keep it super simple and dumb like this:

```
if batch_size is None:
    # original logic
    request = Request(...)
else:
    # new logic
    results = []
    ....
    return results
```

tests/test_document.py
```diff
@@ -1832,6 +1832,16 @@ def test_document_import_bulk(col, bad_col, docs):
...
 
+    # Test import bulk with batch_size
+    results = col.import_bulk(docs, batch_size=len(docs) // 2)
+    assert type(results) is list
+    assert len(results) == 2
+    empty_collection(col)
+
+    result = col.import_bulk(docs, batch_size=len(docs) * 2)
+    assert type(result) is dict
+    empty_collection(col)
```
joowani, line 1843: We should always return a list if `batch_size` is set to a non-None value.

arango/collection.py
aMahanna: We _could_ do it like that, but [f722eff](https://github.com/ArangoDB-Community/python-arango/pull/207/commits/f722effe4fa8136f49c72c47ab6e328d93cc82d0) accomplishes the same goal without duplicating any `Request` & `self._execute` references.

What do you think? Happy to switch to what was recommended if preferred

joowani: I don't want to duplicate either, but I think this is one of those cases where trying to save lines of code results in more bug-prone and inefficient outcome. For example we are initializing the `results` list unnecessarily when `batch_size` is None, and it's harder for new devs to understand what's really happening in line 2032. Also I suspect line 2032 could throw an unhandled KeyError when user passes in an empty documents list.

joowani: I don't want to duplicate either, but I think this is one of those cases where trying to save lines of code results in more bug-prone and inefficient outcome. For example we are initializing the `results` list unnecessarily when `batch_size` is None, and it's harder for new devs to understand what's really happening in line 2032. Also I suspect line 2032 could throw an unhandled KeyError when user passes in an empty documents list.

arango/collection.py
```diff
@@ -2006,21 +2016,25 @@ def import_bulk(
...
+                method="post",
+                endpoint="/_api/import",
+                data=batch,
+                params=params,
+                write=self.name,
+            )
+
+            results.append(self._execute(request, response_handler))
+
+        return results[0] if batch_size is None else results
```
joowani, line 2037: See https://github.com/ArangoDB-Community/python-arango/pull/207#discussion_r904289441.

arango/collection.py
aMahanna: Makes sense, changed as of [c65af2a](https://github.com/ArangoDB-Community/python-arango/pull/207/commits/c65af2add238347c12d990a70bb7fb733a0a0ed7)

arango/collection.py
aMahanna: Addressed in [c65af2a](https://github.com/ArangoDB-Community/python-arango/pull/207/commits/c65af2add238347c12d990a70bb7fb733a0a0ed7)

joowani:
> Some questions
> 
> 1. Does `python-arango`'s readthedocs generate documentation based on anything else but the docstrings? Wondering if I should be aware of any other API documentation that I should be editing accordingly
> 2. Curious to know the reason for having both `_ensure_key_from_id()` and `_ensure_key_in_body()`. Not currently sure

1. Only the API specifications are generated from the docstrings. Take a look at the contents in docs directory for the rest. 
2. Honestly I don't remember anymore haha. The functions do slightly different things, so I guess I wanted to run only the code that was strictly necessary (since they run per document which means a lot).

PR #209, opened 2022-06-28
Prevent batch_size parameter usage when overwrite is set to True in bulk_import  

aMahanna:
Now that `batch_size` has been introduced to `import_bulk()`, we should be checking that users don't use both the `overwrite` and the `batch_size` parameters in the same call.

If server-side batching is ever introduced to `/_api/import`, then this restriction could most likely be lifted.

Warrants a 7.4.1 release

### DIFF
--- a/arango/collection.py
+++ b/arango/collection.py
@@ -1992,12 +1992,16 @@ class Collection(ApiGroup):
             IMPORTANT NOTE: this parameter may go through breaking changes
             in the future where the return type may not be a list of result
             objects anymore. Use it at your own risk, and avoid
-            depending on the return value if possible.
+            depending on the return value if possible. Should not be used
+            if **overwrite** is set to true.
         :type batch_size: int
         :return: Result of the bulk import.
         :rtype: dict | list[dict]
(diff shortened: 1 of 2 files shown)


arango/collection.py
```diff
@@ -1992,12 +1992,16 @@ def import_bulk(
...
             objects anymore. Use it at your own risk, and avoid
-            depending on the return value if possible.
+            depending on the return value if possible. Should not be used
+            if **overwrite** is set to true.
         :type batch_size: int
         :return: Result of the bulk import.
         :rtype: dict | list[dict]
         :raise arango.exceptions.DocumentInsertError: If import fails.
         """
+        if overwrite and batch_size:
```
joowani, line 2002: I think it's better to do `if overwrite and batch_size is not None` to not have to worry about what happens when `batch_size == 0`

joowani (commented):
A couple of nits on the docstrings (I'd like to be picky with documentation as much as I am with code). Thank you for making this change.

joowani (approved):

Issue #210, opened 2022-07-12
Discontinued: /_api/export


aMahanna:
ArangoDB's HTTP Interface for Bulk Exports was previously discontinued in favour of AQL queries and streaming cursors. 

i.e this no longer works
```
POST /_api/export?colleciton=cities
``` 

We should update https://docs.python-arango.com/en/main/specs.html#arango.collection.Collection.export to notify users that this functionality has been discontinued.

aMahanna:
Deprecation announcement: https://www.arangodb.com/docs/stable/release-notes-api-changes39.html#export-api

aMahanna:
internal convo: https://arangodb.slack.com/archives/C1PJT16QP/p1657639563320899

PR #211, opened 2022-08-01
Breaking change: Removed discontinued collection export method

tjoubert:
Solves #210 
- Removed collection.export() method
- Consumers should execute AQL into a `Cursor`


joowani (approved):
Thanks for the PR

PR #213, opened 2022-08-05
Add 'read_timeout' option in client.py

tjoubert:
- Added `request_timeout` property to `ArangoClient` to overwrite `DefaultHTTPClient.REQUEST_TIMEOUT`
- Added `request_timeout` constructor parameter in `ArangoClient` to initialize `request_timeout` property. The default value for this parameter is -1. Any other value supplied, including `None` will overwrite `DefaultHTTPClient.REQUEST_TIMEOUT` 
- Added `# type: ignore` to lines 141 and 146 because they failed `mypy`. However, they work perfectly well when tested. Any suggestions here is very much appreciated.

### DIFF
--- a/arango/client.py
+++ b/arango/client.py
@@ -59,7 +59,7 @@ class ArangoClient:
        request timeout set in the http client.
        None: No timeout.
        int: Timeout value in seconds.
-    :type request_timeout: Optional[int]
+    :type request_timeout: Any
     """
 
     def __init__(
--- a/arango/client.py
+++ b/arango/client.py
@@ -54,9 +54,8 @@ class ArangoClient:
        str: Path to a custom CA bundle file or directory.
     :type verify_override: Union[bool, str, None]
     :param request_timeout: This is the default request timeout (in seconds)
-       for http requests issued by the client. The default value is -1.
-       Setting this parameter to any other value, will overwrite your
-       request timeout set in the http client.
+       for http requests issued by the client if the parameter http_client is
+       not secified. The default value is 60.
        None: No timeout.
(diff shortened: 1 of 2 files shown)


arango/client.py
```diff
@@ -64,6 +71,7 @@ def __init__(
         serializer: Callable[..., str] = lambda x: dumps(x),
         deserializer: Callable[[str], Any] = lambda x: loads(x),
         verify_override: Union[bool, str, None] = None,
+        request_timeout: Any = -1,
```
joowani, line 74: Instead of introducing`-1` I think it might be better to use `60` as default to align with https://github.com/ArangoDB-Community/python-arango/blob/main/arango/http.py#L69

tjoubert: @joowani, thanks for your review. We avoided giving the parameter a real default value (i.e. `60`) because of cases where a caller would first initialize an http client and configure a request timeout there already (let's say `200`) and then pass that object to the constructor of the `ArangoClient`. In this case, the configured request timeout on the custom http client (`200`) will be overwritten by our default value (`60`).

joowani: I see. What about `None`? I would like to avoid introducing a new magic number `-1` to denote a behavior when we've already been using `None` everywhere in the library.

tjoubert: I understand your concern with using `-1`. I have changed the default parameter value to `60` and updated to constructor to overwrite request timeout only if `http_client = None`. This way, it will not overwrite the request timeout on an existing `http_client`.

joowani (approved):
lgtm

PR #220, opened 2022-09-26
3:10 Updated Pregel API Features

tjoubert:
- Added support to retrieve all running jobs 
https://www.arangodb.com/docs/devel/http/pregel.html#get-currently-running-pregel-jobs
- Added the new `details` JSON element in 3.10
https://github.com/arangodb/docs/blob/main/3.10/release-notes-api-changes310.md#pregel-api

### DIFF
--- a/arango/formatter.py
+++ b/arango/formatter.py
@@ -1107,50 +1107,7 @@ def format_pregel_job_data(body: Json) -> Json:
 
     # The detail element was introduced in 3.10
     if "detail" in body:
-        d: Json = {}
-        detail = body["detail"]
-        if "workerStatus" in detail:
-            d["workerStatus"] = detail["workerStatus"]
-        if "aggregatedStatus" in detail:
-            aggregatedStatus = detail["aggregatedStatus"]
-            aStat: Json = {}
-            if "timeStamp" in aggregatedStatus:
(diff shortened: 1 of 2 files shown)


arango/formatter.py
```diff
@@ -1072,26 +1072,105 @@ def format_pregel_job_data(body: Json) -> Json:
...
+                            ri["messagesReceived"] = i["messagesReceived"]
+                        if "memoryBytesUsedForMessages" in i:
+                            ri["memoryBytesUsedForMessages"] = i[
+                                "memoryBytesUsedForMessages"
+                            ]
+                        itemList.append(ri)
+                    agStat["items"] = itemList
+                aStat["allGssStatus"] = agStat
+            d[aggregatedStatus] = aStat
+        result["detail"] = d
```
joowani, line 1153: I'm ok if we want to do just this:
```
if "detail" in body:
    result["detail"] = body["detail"]  # or deepcopy
```
I'll let you decide. Feel free to merge after.

joowani (approved):

arango/formatter.py
tjoubert: I think what you're suggesting is much cleaner. Thanks.

PR #222, opened 2022-09-30
3.10: Read from Followers (allow dirty read)

tjoubert:
Added `allow_dirty_read` optional parameter to methods for the following:
- Single document reads (`GET /_api/document`)
- Batch document reads (`PUT /_api/document?onlyget=true`)
- Read-only AQL queries (`POST /_api/cursor`)
- The edge API (`GET /_api/edges`)
- Read-only Stream Transactions and their sub-operations (`POST /_api/transaction/begin etc.`)

https://github.com/arangodb/docs/blob/main/3.10/release-notes-api-changes310.md#read-from-followers

### DIFF
--- a/arango/aql.py
+++ b/arango/aql.py
@@ -266,7 +266,7 @@ class AQL(ApiGroup):
         skip_inaccessible_cols: Optional[bool] = None,
         max_runtime: Optional[Number] = None,
         fill_block_cache: Optional[bool] = None,
-        allow_dirty_read: Optional[bool] = None,
+        allow_dirty_read: bool = False,
     ) -> Result[Cursor]:
         """Execute the query and return the result cursor.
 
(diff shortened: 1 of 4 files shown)


joowani (commented):
Do we need to make allow_dirty_read syllable? It seems like we can get away with just defaulting to False.

tjoubert:
> Do we need to make allow_dirty_read nullable? It seems like we can get away with just defaulting to False.

Thanks @joowani. I have updated the default value of the allow_dirty_read parameter to False.

arango/aql.py
```diff
@@ -358,6 +359,8 @@ def execute(
             query will not make it into the RocksDB block cache if not already
             in there, thus leaving more room for the actual hot set.
         :type fill_block_cache: bool
+        :param allow_dirty_read: Allow reads from followers in a cluster.
+        :type allow_dirty_read: bool | None
```
joowani, line 363: nit: `allow_dirty_read: bool`

joowani (approved):

PR #228, opened 2022-11-21
Added header for driver version and other info

tjoubert:
@joowani, please check the related issue for context information.
The aim here is to pass this header to every request. So, I've added it to normalize_headers() in request.py.

Please check request.py line 12. Is there a way to quickly retrieve the name and version of the current package in Python?
I am thinking of using the importlib-metadata package (https://pypi.org/project/importlib-metadata/). Do you think that this is a good approach or is there a better way?


arango/request.py
```diff
@@ -2,13 +2,21 @@
 
 from typing import Any, MutableMapping, Optional
 
-from arango.typings import Fields, Headers, Params
+from arango.typings import DriverFlags, Fields, Headers, Params
 
 
-def normalize_headers(headers: Optional[Headers]) -> Headers:
+def normalize_headers(
+    headers: Optional[Headers], driverFlags: Optional[DriverFlags] = None
```
joowani, line 9: nit: `driverFlags` -> `driver_flags` (we won't format the field names in JSON payload and response bodies to be snake_case anymore, but we should still use snake_case for the Python code itself).

arango/request.py
```diff
@@ -2,13 +2,21 @@
...
+) -> Headers:
+    flags = ""
+    if driverFlags is not None:
+        for flag in driverFlags:
+            flags = flags + flag + ";"
+    driverInfo = "python-arango/7.5.2 (" + flags + ")"
```
joowani, line 15: Same as above. `driverInfo` -> `driver_info` (or better yet `driver_header`).

tests/test_request.py
```diff
@@ -11,15 +11,15 @@ def test_request_no_data():
...
-    assert request.headers == {
-        "charset": "utf-8",
-        "content-type": "application/json",
-        "foo": "bar",
-    }
+    # assert request.headers == {
+    #     "charset": "utf-8",
+    #     "content-type": "application/json",
+    #     "foo": "bar",
+    # }
```
joowani, line 18: We can just change these to:
```
assert request.headers["charset"] == "utf-8"
assert request.headers["content-type"] == "application/json"
assert request.headers["foo"] == "bar" 
```

joowani:
> Please check request.py line 12. Is there a way to quickly retrieve the name and version of the current package in Python? I am thinking of using the importlib-metadata package (https://pypi.org/project/importlib-metadata/). Do you think that this is a good approach or is there a better way?

You can simply do:
```
from arango.version import __version__
```
I think it's safe to assume that when we are executing the python-arango code, the version file would've already been generated by `pip install` (`pip install -e` in dev).

arango/request.py
tjoubert: This is fixed.

tests/test_request.py
tjoubert: This is fixed.

tjoubert:
> > Please check request.py line 12. Is there a way to quickly retrieve the name and version of the current package in Python? I am thinking of using the importlib-metadata package (https://pypi.org/project/importlib-metadata/). Do you think that this is a good approach or is there a better way?
> 
> You can simply do:
> 
> ```
> from arango.version import __version__
> ```
> 
> I think it's safe to assume that when we are executing the python-arango code, the version file would've already been generated by `pip install` (`pip install -e` in dev).

@joowani , thanks for the tip. I tested it and it works!

arango/request.py
tjoubert: This is fixed

tests/test_request.py
```diff
@@ -30,15 +30,15 @@ def test_request_string_data():
...
```
joowani, line 37: Looks like you missed this one.

tests/test_request.py
```diff
@@ -68,9 +66,9 @@ def test_request_transaction_data():
...
     assert request.params == {"bool": "1"}
-    assert request.headers == {
-        "charset": "utf-8",
-        "content-type": "application/json",
-        "foo": "bar",
-    }
+    # assert request.headers == {
+    #     "charset": "utf-8",
+    #     "content-type": "application/json",
+    #     "foo": "bar",
```
joowani, line 72: This one also.

arango/request.py
```diff
@@ -86,6 +99,7 @@ class Request:
         "write",
         "exclusive",
         "deserialize",
+        "driverFlags",
```
joowani, line 102: driverFlags -> driver_flags

joowani (approved):
Looks like you missed a couple of spots but the rest looks good! Please feel free to merge after you address the nits. Thanks as always.

tests/test_request.py
tjoubert: Fixed

tests/test_request.py
tjoubert: Fixed

arango/request.py
tjoubert: Fixed

Issue #231, opened 2022-11-29
API Specification Docs not loading


cw00dw0rd:
API Spec information is no longer loading in the documentation: [https://docs.python-arango.com/en/main/specs.html](https://docs.python-arango.com/en/main/specs.html)

joowani:
For now the dev one works: https://docs.python-arango.com/en/dev/

tjoubert:
@joowani , do I understand correctly that `version.py` is generated by the tool `setuptools_scm`? If yes, is it possible to get this tool to write the version number in a text/config file somewhere that we can read from the code? This new file must become part of the deployed package when the driver is referenced/import by other projects.

tjoubert:
@joowani , `setuptools_scm` says this about retrieving package version at runtime:
https://github.com/pypa/setuptools_scm#retrieving-package-version-at-runtime
Looks like we'll have to consider using `importlib.metadata`. What do you think?

joowani:
The link you provided actually covers sphinx (our doc) also so I think we can just follow that: https://github.com/pypa/setuptools_scm#usage-from-sphinx

tjoubert:
@joowani , so we have to do this:
```
from importlib_metadata import version, PackageNotFoundError

name = "python-arango"
driver_version = version(name)
```

Do you think it will work referencing the `python-arango` inside that package itself? The documentation seems to be referring to external packages that have been referenced/imported.

joowani:
I'm not sure but you can test it out by running:

```
python -m sphinx -b html -W docs docs/_build
```

PR #232, opened 2022-11-30
Removed reference to arango.version in request.py

tjoubert:
This is necessary to remove the docs errors while we find a solution for driver version.


tjoubert:
This PR solves #231

joowani (approved):

PR #233, opened 2022-11-30
Added importlib.metadata for version check

tjoubert:
This permanently solves issue #231 while enabling us to pull the version of the driver in the Request file.

### DIFF
--- a/arango/request.py
+++ b/arango/request.py
@@ -28,7 +28,7 @@ def generate_driver_header(driver_flags: Optional[DriverFlags] = None) -> str:
         for flag in driver_flags:
             flags = flags + flag + ";"
     name: str = "python-arango"
-    driver_version: str = version(name)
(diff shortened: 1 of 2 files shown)


arango/request.py
```diff
@@ -26,6 +22,18 @@ def normalize_headers(
...
+def generate_driver_header(driver_flags: Optional[DriverFlags] = None) -> str:
+    flags: str = ""
+    if driver_flags is not None:
+        for flag in driver_flags:
+            flags = flags + flag + ";"
+    name: str = "python-arango"
+    driver_version: str = version(name)
+    driver_header: str = name + "/" + driver_version + " (" + flags + ")"
+
+    return driver_header
```
joowani, line 34: We have to make sure this doesn't break when we do `pip install -e .` for development mode.

PR #237, opened 2022-12-13
Support overload control

tjoubert:
Added `OverloadControlDatabase`, enabling the client to react effectively to potential server overloads.

Requests go through the `OverloadControlApiExecutor`. For every API call, it allows clients to pass the `x-arango-max-queue-time-seconds` request header and also return the `X-Arango-Queue-Time-Seconds` response header if it exists.

Additional tests and documentation added :heavy_check_mark:


joowani (commented):
The approach makes sense. Regarding the name `QueueTimeApiExecutor`, perhaps we could rename it to something more generic so we can reuse to support other headers in the future (e.g. ApiExecutorV2)

CryptoNinjaGeek (approved):

PR #247, opened 2023-05-19
[DE-593] Support latest ArangoDB version

apetenchea:
### Scope & Purpose

Currently, the `main` branch is only tested against ArangoDB 3.7 single-server.
This PR bumps that version to 3.10.6. Due to that, additional changes were needed:
- make sure newly (since 3.7) are available
- adapt unit tests (some were broken from before, some became broken due to newer ArangoDB versions)
- decide the officially supported Python versions (we're going to support the default Python versions that come with the last two Ubuntu LTS releases)

Apart from bumping the ArangoDB version to 3.10.6, the CI will stay the same, until we decide regarding the _CircleCI_ integration. Until then, I am using the `tester.sh` script locally for running more comprehensive suites. I ran **everything** against 3.9.10 and 3.10.6. Rest assured, this is not going to stay like that. I will push for a proper ~~Circle~~CI.

Once this PR gets through, the plan is to add support for 3.11.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature
- [ ] :fire: Performance improvement
- [x] :hammer: Refactoring/simplification

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -13,11 +13,11 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
 
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11.1"]
+        python-version: ["3.8", "3.10"]
 
     steps:
       - name: Checkout repository
@@ -28,14 +28,8 @@ jobs:
 
       - name: Create ArangoDB Docker container
         run: >
-          docker create --name arango -p 8529:8529 -e ARANGO_ROOT_PASSWORD=passwd
-          arangodb/arangodb:3.7.7 --server.jwt-secret-keyfile=/tmp/keyfile
-
-      - name: Copy Foxx service zip into ArangoDB Docker container
-        run: docker cp tests/static/service.zip arango:/tmp/service.zip
-
-      - name: Copy keyfile into ArangoDB Docker container
-        run: docker cp tests/static/keyfile arango:/tmp/keyfile
+          docker create --name arango -p 8529:8529 -e ARANGO_ROOT_PASSWORD=passwd -v "$(pwd)/tests/static/":/tests/static
+          arangodb/arangodb:3.10.6 --server.jwt-secret-keyfile=/tests/static/keyfile
 
       - name: Start ArangoDB Docker container
         run: docker start arango
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -17,7 +17,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.8", "3.10"]
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
 
     steps:
       - name: Checkout repository
--- a/arango/formatter.py
+++ b/arango/formatter.py
@@ -65,6 +65,8 @@ def format_index(body: Json) -> Json:
         result["cacheEnabled"] = body["cacheEnabled"]
     if "legacyPolygons" in body:
         result["legacyPolygons"] = body["legacyPolygons"]
+    if "estimates" in body:
+        result["estimates"] = body["estimates"]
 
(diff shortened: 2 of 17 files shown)


arango/formatter.py
```diff
@@ -215,18 +217,24 @@ def format_collection(body: Json) -> Json:
...
-                "name": cv["name"],
-                "expression": cv["expression"],
-                "overwrite": cv["overwrite"],
-                "computedOn": cv["computedOn"],
-                "keepNull": cv["keepNull"],
-                "failOnWarning": cv["failOnWarning"],
-            }
-            for cv in body["computedValues"]
-        ]
+    if "computedValues" in body:
```
geenen124, line 220: This diff changes the function's default behavior. Before, `computedValues` was only present in `results` if `body["computedValues"]`'s value was non-null. A user could assume that `computedValues` is non-null if it's present in `results`, so this change is breaking.

Unless we're planning a major version release next, IMO we should keep the old behavior instead. What do you think?

.github/workflows/build.yaml
```diff
@@ -13,11 +13,11 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
 
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11.1"]
+        python-version: ["3.8", "3.10"]
```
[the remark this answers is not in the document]
apetenchea: Sure, I just added them back!

Let me explain my intention:
After discussing with Carsten, it turns out our goal is to focus our support and testing efforts on the Python versions that come with the last two Ubuntu LTS versions (3.8 and 3.10), as these are the most commonly used in our user base.
The current github action serving as CI is **only spawning a single-server community instance and runs the single-server driver tests**.
Eventually, to do it properly, we'll have to tests it against:
- single server community
- single server enterprise
- cluster community
- cluster enterprise
The officially supported ArangoDB versions are 3.9, 3.10 and 3.11 (soon). That gives us 12 combinations. We could think of ways of shrinking it down to 8 or so, but it's not a short list. Adding the Python version as the "third dimension" of this matrix will make things even more complicated. Nevertheless, we will ensure that at least the officially supported ones are tested. I could gather a reduced test-set that can be used for this purpose: validating the driver against the given Python version.

Nevertheless, for now it doesn't hurt to leave 3.9 and 3.11 in there.

arango/formatter.py
apetenchea: Thanks for pointing that out! This change was made to make the tests pass, which seemed to have been failing before, because we would always expect `computedValues` to appear in the `result`, as long as it appears in the `body`.
For some reason, it was merged with the failing tests. I assume it was due to the fact that the *CI never checked 3.10 before*, it was only validated on 3.7 single-server.
It's possible that the tests were written with the wrong assumptions or the function's behavior isn't well-defined when it comes to null values in `computedValues`.
1. If we're confident that the tests are correct and reflect the intended behavior of our code, we can accept this change. Being new to the project, I used the tests as guidance, so this was more of a bug-fix to me. Note that for all other keys, their value will be reflected in `result`, as long as they are `in` the `body`. This was the only place where I could find this sort of check. It can be `null` in the JSON response, which translates to `None` in `body`, and the tests expect to see the same in `result`.
2. It is also possible that the tests were never adapted to the newly introduced keys. In that case, we should revisit them and align them with the function's current behavior.
What are your thoughts on this?

apetenchea: I followed with your suggestion and reverted the way we calculate `computedValues` back to the old behavior.
It is unclear whether or not there should be `None` values in the driver's result. Some attributes are simply copied over to the result, while others, such as `computedValues`, are only processed if they are not `None`. In order to adapt the tests, I had to remove `None` values from both the `result` and the `body` before comparing them. There is a common function used to validate the differences between the request's result body and the expected result after it being processed by the driver: _mock_verify_format_

CryptoNinjaGeek (approved):

geenen124 (approved):

Issue #249, opened 2023-05-25
Replacement for `pkg_resources`


jbylund:
`pkg_resources` is now deprecated in favor of `importlib.metadata` / `importlib_metadata` ([source](https://setuptools.pypa.io/en/latest/pkg_resources.html)). Right now it causes warnings and in the future (3.12?) I think it will stop working entirely. I saw there was a related [pr](https://github.com/ArangoDB-Community/python-arango/pull/233) but it was using `importlib.metadata` which apparently has different behaviors on some versions of python and caused issues for some users ([ref](https://github.com/ArangoDB-Community/python-arango/issues/244)). Maybe it would be acceptable to use `importlib_metadata` (the backport) either on all versions of python or only when necessary.

apetenchea:
Fixed in https://github.com/ArangoDB-Community/python-arango/pull/250

Issue #244, opened 2023-02-06
python-arango is not working without pip-install installation


alexbakharew:
Hello!

I am uisng `python-arango` for benchmarking arangodb and elasticsearch. `python-arango` is not installed on the target machine but I am importing it in the following way:
```
# avoid using 'pip install arango-python' command
arango_path = os.getcwd() + "/../iresearch.deps/python-arango"
if arango_path not in sys.path:
    sys.path.insert(0, arango_path)
```
I followed the discussion in this [issue](https://github.com/ArangoDB-Community/python-arango/issues/231) and found following [PR](https://github.com/ArangoDB-Community/python-arango/pull/233) which broke our existing testing pipeline with this error message:

```
alexey@c4:/home/jenkins/workspace/arangodb-ANY-benchmarks-client/arangosearch-benchmarks$ python3 ./Benchmarks.py ../args.json arango
/usr/lib/python3/dist-packages/requests/__init__.py:78: RequestsDependencyWarning: urllib3 (1.26.14) or chardet (3.0.4) doesn't match a supported version!
  warnings.warn("urllib3 ({0}) or chardet ({1}) doesn't match a supported "
Traceback (most recent call last):
  File "/home/jenkins/workspace/arangodb-ANY-benchmarks-client/arangosearch-benchmarks/./Benchmarks.py", line 182, in <module>
    main()
  File "/home/jenkins/workspace/arangodb-ANY-benchmarks-client/arangosearch-benchmarks/./Benchmarks.py", line 128, in main
    ArangoInstance.prepare_environment(environment_config["arango"], engines["arango"]["endpoints"])
  File "/home/jenkins/workspace/arangodb-ANY-benchmarks-client/arangosearch-benchmarks/Arango.py", line 169, in prepare_environment
    remove_all_views(db)
  File "/home/jenkins/workspace/arangodb-ANY-benchmarks-client/arangosearch-benchmarks/Arango.py", line 109, in remove_all_views
    for view in db.views():
  File "/home/jenkins/workspace/arangodb-ANY-benchmarks-client/arangosearch-benchmarks/../iresearch.deps/python-arango/arango/database.py", line 2092, in views
    request = Request(method="get", endpoint="/_api/view")
  File "/home/jenkins/workspace/arangodb-ANY-benchmarks-client/arangosearch-benchmarks/../iresearch.deps/python-arango/arango/request.py", line 128, in __init__
    self.headers: Headers = normalize_headers(headers, driver_flags)
[...]

PR #250, opened 2023-05-25
Replace pkg_resources with importlib_metadata

jbylund:
Copying description from [issue](https://github.com/ArangoDB-Community/python-arango/issues/249):

`pkg_resources` is now deprecated in favor of `importlib.metadata` / `importlib_metadata` ([source](https://setuptools.pypa.io/en/latest/pkg_resources.html)). Right now it causes warnings and in the future (3.12?) I think it will stop working entirely. I saw there was a related [pr](https://github.com/ArangoDB-Community/python-arango/pull/233) but it was using `importlib.metadata` which apparently has different behaviors on some versions of python and caused issues for some users ([ref](https://github.com/ArangoDB-Community/python-arango/issues/244)). Maybe it would be acceptable to use `importlib_metadata` (the backport) either on all versions of python or only when necessary.

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -1,8 +1,8 @@
 import json
 from typing import Union
 
+import importlib_metadata
 import pytest
-from pkg_resources import get_distribution
 from requests import Session
 
 from arango.client import ArangoClient
@@ -21,7 +21,7 @@ def test_client_attributes():
     http_client = DefaultHTTPClient()
 
     client = ArangoClient(hosts="http://127.0.0.1:8529", http_client=http_client)
-    assert client.version == get_distribution("python-arango").version
+    assert client.version == importlib_metadata.version("python-arango")
     assert client.hosts == ["http://127.0.0.1:8529"]
 
     assert repr(client) == "<ArangoClient http://127.0.0.1:8529>"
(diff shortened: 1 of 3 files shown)


tests/test_client.py
```diff
@@ -2,8 +2,8 @@
 from typing import Union
 
 import pytest
-from pkg_resources import get_distribution
 from requests import Session
+import importlib_metadata
```
apetenchea, line 6: Thanks for spotting that!

I would be ready to accept this PR, but the pre-commit hook `isort` is failing due to the non-standard ordering of imports. Please see [CONTRIBUTING.md](https://github.com/ArangoDB-Community/python-arango/blob/main/CONTRIBUTING.md) for more details.

apetenchea (changes requested):

jbylund:
I'm a little worried about the use case of #244 (using without installing via just adding to path). But it seems to work for me:
```
In [5]: import arango

In [6]: arango.__file__
Out[6]: '/Users/joebylund/scratch/python-arango/arango/__init__.py'

In [7]: ArangoClient().version
Out[7]: '7.5.8.dev2+gfac52f3'
```

apetenchea:
> I'm a little worried about the use case of #244 (using without installing via just adding to path). But it seems to work for me:
> 
> ```
> In [5]: import arango
> 
> In [6]: arango.__file__
> Out[6]: '/Users/joebylund/scratch/python-arango/arango/__init__.py'
> 
> In [7]: ArangoClient().version
> Out[7]: '7.5.8.dev2+gfac52f3'
> ```

That's a valid concern, I'll have to test it out as well. Will get to it by the end of this week.

apetenchea (approved):
I tried it locally as well and it works for me. However, I am hesitant to merge it yet, without @alexbakharew 's input, as this can potentially affect some of our internal projects.

apetenchea:
Alright, I talked to @alexbakharew and he confirmed it's working.
Thank you for you contribution!

PR #252, opened 2023-06-06
Improvements

alexbakharew:
Hello  @joowani @apetenchea !

I decided to create this PR because I really need this changes for my Benchmarks project. Here is the short description of changes:

- Added 'details' boolean parameter for getting more detailed version description: `/_db/_system/_api/version?details=true`
- Remove only service fields from view definitions. It will help us in future when new features will be added. I think we should do the same thing for every DDL objects
 
 What do you think about this PR?

### DIFF
--- a/arango/formatter.py
+++ b/arango/formatter.py
@@ -854,44 +854,17 @@ def format_view(body: Json) -> Json:
     :return: Formatted body.
     :rtype: dict
     """
-    result: Json = {}
+    # Remove only unnecessary keys
+    if "error" in body:
+        body.pop("error")
+    if "code" in body:
+        body.pop("code")
     if "globallyUniqueId" in body:
-        result["global_id"] = body["globallyUniqueId"]
-    if "id" in body:
-        result["id"] = body["id"]
-    if "name" in body:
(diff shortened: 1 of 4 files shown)


arango/formatter.py
```diff
@@ -854,44 +854,17 @@ def format_view(body: Json) -> Json:
...
-    if "name" in body:
-        result["name"] = body["name"]
-    if "type" in body:
-        result["type"] = body["type"]
-    if "cleanupIntervalStep" in body:
-        result["cleanup_interval_step"] = body["cleanupIntervalStep"]
-    if "commitIntervalMsec" in body:
-        result["commit_interval_msec"] = body["commitIntervalMsec"]
-    if "consolidationIntervalMsec" in body:
-        result["consolidation_interval_msec"] = body["consolidationIntervalMsec"]
```
apetenchea, line 871: Unfortunately, we can't go with this change, because it affects backwards compatibility. For example, users would expect to find `consolidation_interval_msec` instead of `consolidationIntervalMsec`. Unless we do a major release, IMO we can't do it this way.

apetenchea (approved):

PR #257, opened 2023-07-12
[BTS-1515] Pass PoolTimeout as argument

apetenchea:
This PR makes the `DefaultHTTPClient` more flexible, adding the following new parameters:
- request_timeout: socket timeout in seconds for each individual connection
- pool_connections: the number of urllib3 connection pools to cache
- pool_maxsize: the maximum number of connections to save in the pool
- pool_timeout: if set, then the pool will be set to block=True, and requests will block for pool_timeout seconds and raise `EmptyPoolError` if no connection is available within the time period

The ones prefixed with *_pool* are all passed to the `PoolManager`. Note that these values cannot be changed once the HTTP client has been created.

The following constants were already used before, but are now configurable as parameters:
- request_timeout 
- retry_attempts
- backoff_factor

Note that `request_timeout` is mapped to the parameter with the same name of the HTTP client. Unlike the other parameters, this one can be changed.

The default values stay the same. If clients don't specify any parameters, nothing changes.

**Example**
```python
http_client = DefaultHTTPClient(
    request_timeout=80,
    retry_attempts=5,
    backoff_factor=1.0,
    pool_connections=16,
    pool_maxsize=12,
    pool_timeout=120,
)
client = ArangoClient(hosts="http://127.0.0.1:8529", http_client=http_client)
db = client.db(db.name, username, password, verify=True)
client.request_timeout = 100  # increase timeout for all requests
```

### DIFF
--- a/arango/http.py
+++ b/arango/http.py
@@ -117,7 +117,7 @@ class DefaultHTTPClient(HTTPClient):
     :param retry_attempts: Number of retry attempts.
     :type retry_attempts: int
     :param backoff_factor: Backoff factor for retry attempts.
-    :type backoff_factor: int
+    :type backoff_factor: float
     :param pool_connections: The number of urllib3 connection pools to cache.
     :type pool_connections: int
     :param pool_maxsize: The maximum number of connections to save in the pool.
@@ -130,12 +130,12 @@ class DefaultHTTPClient(HTTPClient):
         self,
         request_timeout: int = 60,
         retry_attempts: int = 3,
-        backoff_factor: int = 1,
+        backoff_factor: float = 1.0,
         pool_connections: int = 10,
         pool_maxsize: int = 10,
         pool_timeout: Union[int, float, None] = DEFAULT_POOL_TIMEOUT,
     ) -> None:
-        self._request_timeout = request_timeout
+        self.request_timeout = request_timeout
         self._retry_attempts = retry_attempts
         self._backoff_factor = backoff_factor
         self._pool_connections = pool_connections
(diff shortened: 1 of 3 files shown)


arango/http.py
```diff
@@ -63,12 +70,77 @@ def send_request(
...
+    :type pool_timeout: int | float | None
+    :param kwargs: Additional keyword arguments passed to the HTTPAdapter constructor.
+    :type kwargs: Any
+    """
+
+    def __init__(
+        self,
+        pool_connections: int = DEFAULT_POOLSIZE,
+        pool_maxsize: int = DEFAULT_POOLSIZE,
+        pool_timeout: Union[int, float, None] = DEFAULT_POOL_TIMEOUT,
```
jmb-altana, line 90: This is confusingly named. There is an argument to urlopen called `pool_timeout` which is the timeout to wait for a connection from the pool and only matters if `blocking=True`. This has nothing to do with the actual connections, just whether requests block to get a cached connection or just make a new one if all cached connections are busy.

You're mapping this to `timeout` which is a ConnectionPool argument that sets the default connect/read timeout for each connection. But because you set `timeout` on each request, this pool level default will always be ignored

arango/http.py
```diff
@@ -128,7 +205,7 @@ def send_request(
             data=data,
             headers=headers,
             auth=auth,
-            timeout=self.REQUEST_TIMEOUT,
+            timeout=self.request_timeout,
```
jmb-altana, line 208: the result of this is that `pool_timeout` which maps to `timeout` on the connection pool, will always be ignored https://urllib3.readthedocs.io/en/stable/reference/urllib3.connectionpool.html#urllib3.HTTPConnectionPool.urlopen

CryptoNinjaGeek (approved):

PR #259, opened 2023-07-14
[DE-562] Index Cache Refilling

apetenchea:
Adding new per-operation option `refillIndexCache` to write operations:
 
  - single-document write operations (insert, replace, update, delete)
  - multi-document write operations (insert_many, replace_many, update_many, delete_many)

  If the option is set to `True`, new entries are added to in-memory index caches if
  document operations affect the edge index or cache-enabled persistent indexes. Every
  currently running transaction will keep track of which in-memory index cache entries
  were invalidated by the transaction, and will try to (re-)fill them later.

  Example:
  ```python
  collection.insert({"foo": "bar"}, refillIndexCaches=True)
  db.aql.execute("INSERT {foo: bar} INTO collection OPTIONS { refillIndexCaches: true }")
  ```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,25 @@
 main
 ----
 
+* Added new per-operation option `refillIndexCache` to write operations:
+
+  - single-document write operations (insert, replace, update, delete)
+  - multi-document write operations (insert_many, replace_many, update_many, delete_many)
+
+  If the option is set to `True`, new entries are added to in-memory index caches if
+  document operations affect the edge index or cache-enabled persistent indexes. Every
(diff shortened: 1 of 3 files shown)


dothebart (approved):
I would mention Arangodb along with these version number comments, else LGTM.

aMahanna:
> I would mention Arangodb along with these version number comments, else LGTM.

updated as of https://github.com/ArangoDB-Community/python-arango/pull/259/commits/7c2bc15f6317bd6400dbca0498a276bd9cd94d97

feel free to merge

dothebart (approved):

PR #260, opened 2023-07-14
[DE-545] UTF-8 names

apetenchea:
It is the responsibility of the server to enable extended names. The driver can send any UTF-8 string along with the request.
Therefore, this PR doesn't really add any new and exciting features. Instead, it adds two tests (one for databases and one for collections), making sure the driver can successfully forward UTF-8 names to the server.
Note that, in order to run the tests, the server has to be started with the `--database.extended-names` option. Unfortunately, this is only possible starting with 3.11 docker images, and we're currently using 3.10.9 for our GitHub actions. Nevertheless, I have added the parameter as a comment for now. I can confirm the tests have passed locally for versions >= 3.11.0.

### DIFF
--- a/tests/static/cluster.conf
+++ b/tests/static/cluster.conf
@@ -8,4 +8,6 @@ jwt-secret = /tests/static/keyfile
 
 [args]
 all.database.password = passwd
+# Extended names can be used starting with 3.11
+# all.database.extended-names = true
 all.log.api-enabled = true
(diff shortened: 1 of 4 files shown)


dothebart:
please try to create indices on these collections / databases as well.

dothebart:
the test should create indices on database/collections with unicode names.

dothebart (approved):
don't have any opinion on the location of the test. 
LGTM.

PR #262, opened 2023-07-21
[DE-554] `serverId` query param (GET/PUT `/_admin/log/level`)

aMahanna:
Introduced in: `v3.10.2`

The `GET /_admin/log/level` and `PUT /_admin/log/level` endpoints support a new query parameter `serverId`, to forward log level get and set requests to a specific server.

This makes it easier to adjust the log levels in clusters because DB-Servers require JWT authentication whereas Coordinators also support authentication using usernames and passwords.

Introduced: https://github.com/arangodb/docs/pull/1209 
Documentation: https://www.arangodb.com/docs/devel/release-notes-api-changes310.html#log-level-api

### DIFF
--- a/arango/database.py
+++ b/arango/database.py
@@ -572,13 +572,22 @@ class Database(ApiGroup):
 
         return self._execute(request, response_handler)
 
-    def log_levels(self) -> Result[Json]:
+    def log_levels(self, server_id: Optional[str] = None) -> Result[Json]:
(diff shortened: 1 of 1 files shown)


aMahanna:
@dothebart will also have to trouble you with a review on this PR when you get the chance

dothebart (approved):

PR #224, opened 2022-10-13
Only set computedValues if value exists and is not None

153957:
If `computedValues` is `None` it can not be iterated over.

Fixes #223

PR #268, opened 2023-08-08
Computed Values support for `/_api/collection/collection/properties`

aMahanna:
- Introduces a `computer_values` parameter for the `collection.configure` method
- Test cases for setting & fetching [computedValues](https://www.arangodb.com/docs/3.11/data-modeling-documents-computed-values.html)
- discussion required: rework `computerValues` logic in the `format.format_collection` method to support unspecified values in a computedValue dictionary

### DIFF
--- a/arango/collection.py
+++ b/arango/collection.py
@@ -306,6 +306,7 @@ class Collection(ApiGroup):
         schema: Optional[Json] = None,
         replication_factor: Optional[int] = None,
         write_concern: Optional[int] = None,
+        computed_values: Optional[List[Json]] = None
     ) -> Result[Json]:
         """Configure collection properties.
 
@@ -341,6 +342,8 @@ class Collection(ApiGroup):
             data["replicationFactor"] = replication_factor
         if write_concern is not None:
             data["writeConcern"] = write_concern
+        if computed_values is not None:
+            data["computedValues"] = computed_values
 
         request = Request(
             method="put",
--- a/arango/formatter.py
+++ b/arango/formatter.py
@@ -249,20 +249,9 @@ def format_collection(body: Json) -> Json:
         result["min_revision"] = body["minRevision"]
     if "schema" in body:
         result["schema"] = body["schema"]
+    if "computedValues" in body:
+        result["computedValues"] = body["computedValues"]
 
-    # New in 3.10
-    if body.get("computedValues") is not None:
-        result["computedValues"] = [
-            {
-                "name": cv["name"],
-                "expression": cv["expression"],
-                "overwrite": cv["overwrite"],
-                "computedOn": cv["computedOn"],
-                "keepNull": cv["keepNull"],
-                "failOnWarning": cv["failOnWarning"],
-            }
-            for cv in body["computedValues"]
-        ]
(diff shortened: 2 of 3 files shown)


arango/collection.py
```diff
@@ -306,6 +306,7 @@ def configure(
         schema: Optional[Json] = None,
         replication_factor: Optional[int] = None,
         write_concern: Optional[int] = None,
+        computed_values: Optional[List[Json]] = None
```
aMahanna, line 309: small nitpick: we have `computed_values` as the parameter name here, but `computedValues` for the `create_collection` method in `database.py`

https://github.com/ArangoDB-Community/python-arango/blob/main/arango/database.py#L977

arango/formatter.py
```diff
@@ -249,20 +249,9 @@ def format_collection(body: Json) -> Json:
...
-            {
-                "name": cv["name"],
-                "expression": cv["expression"],
-                "overwrite": cv["overwrite"],
-                "computedOn": cv["computedOn"],
-                "keepNull": cv["keepNull"],
-                "failOnWarning": cv["failOnWarning"],
-            }
-            for cv in body["computedValues"]
-        ]
```
aMahanna, line 265: Need to revisit this as I am aware of https://github.com/ArangoDB-Community/python-arango/pull/247#discussion_r1200342072

Building `result["computedValues"]` via a list comprehension will cause `KeyError`s if some optional computedValue properties have not been specified (e.g `computeOn`, `keepNull`, `failOnWarning`)

aMahanna: It seems like https://github.com/ArangoDB-Community/python-arango/pull/224 was introduced as a consequent of using a list comprehension for `computedValues`

Does it not make sense to just follow how the [schema property is set](https://github.com/ArangoDB-Community/python-arango/blob/main/arango/formatter.py#L250-L251)?

arango/collection.py
```diff
@@ -306,6 +306,7 @@ def configure(
         schema: Optional[Json] = None,
         replication_factor: Optional[int] = None,
         write_concern: Optional[int] = None,
+        computed_values: Optional[Jsons] = None,
```
aMahanna, line 309: small nitpick: we have `computed_values` as the parameter name here, but have `computedValues` for the `create_collection` method in `database.py`:

https://github.com/ArangoDB-Community/python-arango/blob/main/arango/database.py#L977

arango/formatter.py
apetenchea: This should be adapted accordingly, especially since the current implementation is prone to exceptions, as you pointed out.

I think it makes perfect sense to follow how the schema property is set. Maybe we should keep the _is not None_, such that _result["computedValues"]_ stays non-null. I imagine something like this:
```python
if body.get("computedValues") is not None:
    result["computedValues"] = body["computedValues"]
```

This should give us the equivalent, without raising _KeyError_. I am not adamant about using _not None_ instead of _"computedValues" in body_, but I am suggesting the former because it does not change the function's behavior.

aMahanna: understood, and I agree that it's probably best not to change the existing behaviour

added in 3acd627

PR #271, opened 2023-08-15
Fix ReadTheDocs Builds

### LINKED ISSUE #272: API Specification Page is not rendering
See https://docs.python-arango.com/en/main/specs.html
Whereas it should look like https://docs.python-arango.com/en/dev/specs.html

- The ReadTheDocs builds have been silently failing for some time now: https://readthedocs.org/projects/python-driver-for-arangodb/builds/
- Possibly related to the fact that the workflow is still using `Python 3.7` (EOL in June)

ReadTheDocs has moved on to a different configuration setup, which requires all ReadTheDocs users to migrate:
```
Your builds will stop working soon!
Configuration files will soon be required by projects, and will no longer be optional. Read our blog post to create one and ensure your project continues building successfully: https://blog.readthedocs.com/migrate-configuration-v2/
```

#271 attempts to address this issue

aMahanna:
fixes #272

Address the ReadTheDocs migration required as indicated by recent builds (which have also been silently failing):

<details>
<summary>Image</summary>

Image

</details>

This is done by introducing a `.readthedocs.yaml` file

See https://blog.readthedocs.com/migrate-configuration-v2/

Preview the fixed API Spec page: https://docs.python-arango.com/en/fix-readthedocs/specs.html


docs/requirements.txt
```diff
@@ -0,0 +1,3 @@
+requests_toolbelt
+importlib_metadata
+jwt
```
aMahanna, line 3: This is required in order for the sphinx `autodoc` functionality to work (for example, see the following `no module named 'jwt'` error: https://readthedocs.org/projects/python-driver-for-arangodb/builds/21620433/)

apetenchea (approved):

PR #273, opened 2023-08-16
new: `raise_on_document_error` param for `collection.update_many`

aMahanna:
New `raise_on_document_error` parameter for `update_many`: 

```
:param raise_on_document_error: Whether to raise if a DocumentRevisionError
            or a DocumentUpdateError is encountered on an individual document,
            as opposed to returning the error as an object in the result list.
            Defaults to False.
```

```py
col.update_many(docs, raise_on_document_error=True)
```

### DIFF
--- a/arango/collection.py
+++ b/arango/collection.py
@@ -1590,6 +1590,7 @@ class Collection(ApiGroup):
         sync: Optional[bool] = None,
         silent: bool = False,
         refill_index_caches: Optional[bool] = None,
+        raise_on_document_error: bool = False,
     ) -> Result[Union[bool, List[Union[Json, ArangoServerError]]]]:
         """Update multiple documents.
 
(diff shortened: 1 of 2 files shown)


arango/collection.py
```diff
@@ -1590,6 +1590,7 @@ def update_many(
         sync: Optional[bool] = None,
         silent: bool = False,
         refill_index_caches: Optional[bool] = None,
+        raise_on_document_error: bool = False,
```
aMahanna, line 1593: Not sure whether to use `Optional[bool] = None` or `bool = False` here, given the mix of default values for some of the parameters in this function. Could use a second opinion

geenen124: Since one of the two will happen a `None` doesn't really have any meaning? A boolean type makes sense here imo

geenen124 (approved):
LGTM

PR #275, opened 2023-08-18
[DE-584] Refactor Deprecated `/_api/simple` methods 

aMahanna:
**Main changes**

- Re-implements the following methods via AQL:
    - `collection.ids`
    -  `collection.keys`
    -  `collection.all`
    -  `collection.find`
    -  `collection.find_in_box` (requesting feedback / second opinion on implementation)
    -  `collection.random`
    -  `collection.update_match`
    -  `collection.replace_match`
    -  `collection.delete_match`

- Implements the `collection.get_index` method (`GET /_api/index/{id}`) as it was previously missing

**Miscellaneous changes**

- New utility function: `build_filter_conditions` (now used in the filter-based simple methods)
- Add `allow_dirty_read` parameter to the reworked simple methods given that they are now implemented using the `_api/cursor` endpoint
- docstring cleanup
- test case cleanup

### DIFF
--- a/arango/collection.py
+++ b/arango/collection.py
@@ -33,6 +33,7 @@ from arango.exceptions import (
     EdgeListError,
     IndexCreateError,
     IndexDeleteError,
+    IndexGetError,
     IndexListError,
     IndexLoadError,
 )
@@ -42,7 +43,14 @@ from arango.request import Request
 from arango.response import Response
 from arango.result import Result
 from arango.typings import Fields, Headers, Json, Jsons, Params
-from arango.utils import get_batches, get_doc_id, is_none_or_int, is_none_or_str
+from arango.utils import (
+    build_filter_conditions,
+    get_batches,
+    get_doc_id,
+    is_none_or_bool,
+    is_none_or_int,
+    is_none_or_str,
+)
 
 
 class Collection(ApiGroup):
@@ -528,7 +536,7 @@ class Collection(ApiGroup):
             is compared against the revision of target document.
         :type check_rev: bool
         :param allow_dirty_read: Allow reads from followers in a cluster.
-        :type allow_dirty_read: bool | None
+        :type allow_dirty_read: bool
         :return: True if document exists, False otherwise.
         :rtype: bool
         :raise arango.exceptions.DocumentInError: If check fails.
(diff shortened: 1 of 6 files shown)


tests/test_document.py
```diff
@@ -1364,7 +1379,7 @@ def test_document_find_in_box(col, bad_col, geo, cluster):
     result = col.find_in_box(
         latitude1=0, longitude1=0, latitude2=6, longitude2=3, limit=0, index=geo["id"]
     )
-    assert clean_doc(result) == [doc1, doc3]
+    assert clean_doc(result) == []
```
aMahanna, line 1382: Based on `limit = 0` this case should have never returned any data, strange 🤔

arango/collection.py
```diff
@@ -893,26 +942,53 @@ def find_in_box(
...
-            data["skip"] = skip
-        if limit is not None:
-            data["limit"] = limit
+        # TODO: REVISIT `geo_str` logic
+        # https://www.arangodb.com/docs/stable/aql/functions-geo.html#within_rectangle
+        # Initial assumption that attribute names are "latitude" and "longitude"
+        geo_str = "[doc.longitude, doc.latitude]"
         if index is not None:
-            data["geo"] = self._name + "/" + index
+            geo_index = self.get_index(f"{self.name}/{index}")
```
aMahanna, line 952: I am in need of a second opinion on the new implementation of `collection.find_in_box`

Given the deprecation of the `WITHIN_RECTANGLE` AQL function, we are now opting to use the `GEO_CONTAINS` method in combination with the `GEO_POLYGON` method: https://www.arangodb.com/docs/stable/aql/functions-geo.html#within_rectangle

The `GEO_CONTAINS` method takes two parameters: the polygon, and the coordinates of the document.

Given that the attribute names representing the coordinates of the document  can vary on a user-basis, we cannot assume that `doc.longitude` and `doc.latitude` will always exist.

In order to prevent any breaking changes to `collection.find_in_box`, we rely on the existing `index` parameter to fetch the Geo Index's `fields` values to use as the document's coordinate attributes.

Therefore, we assume that the Geo Index's `fields` is an array containing either 2 values at most. Otherwise, a `ValueError` is raised.

Perhaps I am overcomplicating things; happy to move to another solution if we see one

aMahanna:
Seems like the tests are currently failing due to the mismatch in JSON data returned from `GET /_api/index` and `GET /_api/index{id}`: https://github.com/ArangoDB-Community/python-arango/actions/runs/5906308034/job/16022064208?pr=275

i.e `GET /_api/index{id}` does not return the `"selectivityEstimate"` property

arango/collection.py
apetenchea: I agree!
Trying to create a GEO index with more than 2 attributes will throw an exception. IMO, double checking it here (along with the fact that the index is truly "geo") is a good idea.

arango/collection.py
```diff
@@ -893,26 +942,53 @@ def find_in_box(
...
-            "latitude2": latitude2,
-            "longitude2": longitude2,
-        }
-        if skip is not None:
-            data["skip"] = skip
-        if limit is not None:
-            data["limit"] = limit
+        # TODO: REVISIT `geo_str` logic
+        # https://www.arangodb.com/docs/stable/aql/functions-geo.html#within_rectangle
+        # Initial assumption that attribute names are "latitude" and "longitude"
```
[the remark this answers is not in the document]
aMahanna: Sounds good, let me know what you think: 55d559d

apetenchea (approved):
LGTM

apetenchea (approved):

PR #279, opened 2023-09-01
[DE-664] Deprecating Batch API

apetenchea:
https://www.arangodb.com/docs/stable/http/batch-request.html
 
 - The batch API is deprecated since ArangoDB 3.8.0 and will be removed in a future version.
  - The `BatchDatabase` is still available, but it now uses a `TreadPoolExecutor` internally.
  - For backwards compatibility, the `BatchDatabase` uses only one worker thread, essentially sending the requests sequentially. Feel free to set the `max_workers` parameter to a higher value if you want to use multiple threads, but be aware that the requests will be sent in parallel, which may cause problems if you are using transactions.
  - To discourage the use of this API, we now issue a warning when the `BatchDatabase` is used. This warning may be disabled by the user.

  Note that `{"foo": "bar"}` may be inserted after `{"foo": "baz"}` in the following example:
  ```python
  with db.begin_batch_execution(max_workers=2) as batch_db:
      job1 = batch_db.collection.insert({"foo": "bar"})
      job2 = batch_db.collection.insert({"foo": "baz"})
  ```

Documentation has been updated accordingly.

### DIFF
--- a/tests/test_batch.py
+++ b/tests/test_batch.py
@@ -1,15 +1,7 @@
-import json
-
-import mock
 import pytest
 
 from arango.database import BatchDatabase
-from arango.exceptions import (
-    BatchExecuteError,
-    BatchJobResultError,
-    BatchStateError,
-    DocumentInsertError,
-)
+from arango.exceptions import BatchJobResultError, BatchStateError, DocumentInsertError
 from arango.job import BatchJob
 from tests.helpers import clean_doc, extract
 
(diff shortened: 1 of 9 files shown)


tests/test_batch.py
```diff
@@ -144,28 +133,3 @@ def test_batch_job_result_not_ready(db, col, docs):
     assert batch_db.commit() == [job]
     assert len(job.result()) == len(docs)
     assert extract("_key", col.all()) == extract("_key", docs)
-
-
-def test_batch_bad_state(db, col, docs):
```
apetenchea, line 149: Note that this test no longer makes sense with the current implementation, because we don't throw this exception anymore.

aMahanna (approved):
LGTM

PR #281, opened 2023-09-01
[DE-670]: `GET /_api/database/user`

aMahanna:
- Introduces `databases_accesible_to_user()` to return the list of databases currently accessible to the user


apetenchea (approved):
LGTM

PR #282, opened 2023-09-01
[DE-668]: `GET /_api/view/{view_name}`

aMahanna:
- Introduces `view_info()` to retrieve basic information about a View. Not to be confused with the existing `view()` method, which returns the entire properties of a View


apetenchea (approved):
LGTM

PR #286, opened 2023-09-12
[DE-667]: `GET /_api/collection/{name}` & `PUT /_api/collection/{name}/compact`

aMahanna:
* introduces `info()` to fetch some information about a collection
* introduces `compact()` to run a compact operation

### DIFF
--- a/arango/collection.py
+++ b/arango/collection.py
@@ -526,7 +526,6 @@ class Collection(ApiGroup):
 
         def response_handler(resp: Response) -> Json:
             if resp.is_success:
-                breakpoint()
                 return format_collection(resp.body)
             raise CollectionCompactError(resp, request)
 
--- a/arango/collection.py
+++ b/arango/collection.py
@@ -9,7 +9,9 @@ from arango.cursor import Cursor
 from arango.exceptions import (
     ArangoServerError,
     CollectionChecksumError,
+    CollectionCompactError,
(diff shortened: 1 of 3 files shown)


arango/collection.py
```diff
@@ -335,6 +337,26 @@ def response_handler(resp: Response) -> Json:
 
         return self._execute(request, response_handler)
 
+    def info(self) -> Result[Json]:
+        """Return the collection information.
```
aMahanna, line 341: not sure if this method should live in `collection.py` or `database.py`, given the precedent of storing certain  `/_api/collection/{name}` endpoints in `database.py` ([source](https://github.com/ArangoDB-Community/python-arango/blob/main/arango/database.py#L901-L1145))

Currently, UX would look like

```py
db.collection('myCollection').info()
```

Alternatively we could have something like:
```py
db.collection_info('myCollection')
```

I have no preference

apetenchea: I believe the current implementation is more user-friendly. It lets users pass around the collection object without worrying about its name or the database.
```python
def foo(col: Collection):
    col.info()
```
Regarding `view_info()`, unfortunately there's no `View` class.

apetenchea (approved):

arango/collection.py
aMahanna: makes sense, will keep it as is 👍

PR #293, opened 2023-11-03
Authenticate using User JWT-Token

### LINKED ISSUE #238: Authenticate using User JWT-Token
It would be great to support User JWT Token-based authentication for a database connection. The benefit is that the python-arango client would not need to know any username/password credentials, and a user does not need the superuser access that the current `superuser_token` argument supports. This does come with added responsibility for a user authenticating this way since there is no guarantee that the token is still valid.

For example, loading a JWT from environment variables:
```python
from arango import ArangoClient
import os

# Initialize the ArangoDB client.
client = ArangoClient()

# get user JWT
token = os.environ["ARANGODB_USER_JWT"]

# Connect to "test" database as a user using the token.
db = client.db('test', user_token=token)
```

aMahanna:
Closes #238 

Introduces a `user_token` parameter to the `arango.client.ArangoClient.db` method to build a `JwtConnection` object without a `username/password` and without `auth_method="jwt"`

also introduces `set_token` to `arango.connection.JwtConnection` and `arango.connection.JwtSuperuserConnection`

```py
import os
from arango import ArangoClient

token = os.environ["ARANGODB_USER_JWT"]
db = ArangoClient().db(user_token=token)
db = ArangoClient().db(username=..., password=..., user_token=token) # Valid, but ignores `username` and `password` params

db.conn.set_token(token=...) # Valid
db.conn.refresh_token() # Raises `arango.exceptions.JWTRefreshError`, since `username/password` is not set
ArangoClient().db(user_token=expired_token) # Raises `JWTExpiredError`
```

### DIFF
--- a/arango/client.py
+++ b/arango/client.py
@@ -171,6 +171,7 @@ class ArangoClient:
         password: str = "",
         verify: bool = False,
         auth_method: str = "basic",
+        user_token: Optional[str] = None,
         superuser_token: Optional[str] = None,
         verify_certificate: bool = True,
     ) -> StandardDatabase:
@@ -189,6 +190,10 @@ class ArangoClient:
             refreshed automatically using ArangoDB username and password. This
             assumes that the clocks of the server and client are synchronized.
         :type auth_method: str
+        :param user_token: User generated token for user access.
+            If set, parameters **username**, **password** and **auth_method**
+            are ignored. This token is not refreshed automatically.
+        :type user_token: str
         :param superuser_token: User generated token for superuser access.
             If set, parameters **username**, **password** and **auth_method**
             are ignored. This token is not refreshed automatically.
@@ -212,6 +217,17 @@ class ArangoClient:
                 serializer=self._serializer,
                 deserializer=self._deserializer,
                 superuser_token=superuser_token,
+            )
+        elif user_token is not None:
+            connection = JwtConnection(
+                hosts=self._hosts,
+                host_resolver=self._host_resolver,
+                sessions=self._sessions,
+                db_name=name,
+                http_client=self._http,
+                serializer=self._serializer,
+                deserializer=self._deserializer,
+                user_token=user_token,
             )
         elif auth_method.lower() == "basic":
             connection = BasicConnection(
--- a/arango/client.py
+++ b/arango/client.py
@@ -192,7 +192,9 @@ class ArangoClient:
         :type auth_method: str
         :param user_token: User generated token for user access.
             If set, parameters **username**, **password** and **auth_method**
-            are ignored. This token is not refreshed automatically.
+            are ignored. This token is not refreshed automatically. If automatic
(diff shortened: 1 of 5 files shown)


arango/client.py
```diff
@@ -213,6 +218,17 @@ def db(
...
+            connection = JwtConnection(
+                hosts=self._hosts,
+                host_resolver=self._host_resolver,
+                sessions=self._sessions,
+                db_name=name,
+                http_client=self._http,
+                serializer=self._serializer,
+                deserializer=self._deserializer,
+                user_token=user_token,
+            )
```
aMahanna, line 231: Curious whether we want to support the possibility of passing a `user_token` **and** and a `username/password` combination? This way a user could rely on the automatic `db.conn.refresh_token` instead of manually calling `db.conn.set_token`

geenen124: Right now we already have that functionality if `auth_method="jwt"` is passed?

tests/test_auth.py
```diff
@@ -121,8 +130,12 @@ def test_auth_jwt_expiry(client, db_name, root_password, secret):
...
-    # Test correct error on token expiry.
+    # Test correct error on token expiry (superuser).
     db = client.db("_system", superuser_token=expired_token)
     with assert_raises(ServerVersionError) as err:
         db.version()
     assert err.value.error_code == FORBIDDEN
+
+    # Test correct error on token expiry (user).
+    with assert_raises(JWTExpiredError) as err:
+        db = client.db("_system", user_token=expired_token)
```
geenen124, line 141: Could add an additional test for `set_token` here?

arango/client.py
aMahanna: Currently if `user_token` and `username/password` is passed (regardless if `auth_method="jwt"` or not), we create a `JwtConnection` object with only `user_token`.

https://github.com/ArangoDB-Community/python-arango/blob/34aaa138e62dace4469c3da6874adba4625f3ef0/arango/client.py#L210-L257

The quick fix to this would be to pass `username/password` into the `user_token`-based `JwtConnection` so that the user can make use of `conn.refresh_token()`, assuming that the `username/password` info is correct

tests/test_auth.py
aMahanna: good catch! added in [74e1b0b](https://github.com/ArangoDB-Community/python-arango/pull/293/commits/74e1b0bb4b6f826577094d9fb5b0cf186cfb4b86)

arango/client.py
aMahanna: resolving after internal discussion:

- `user_token`-based `JwtConnection` remains
- automatic token refresh will only be offered if `username/password` & `auth_method="jwt"` is set
- manual token setter is available for both `JwtConnection` and `JwtSuperuserConnection` objects

geenen124 (approved):
LGTM!

PR #299, opened 2023-11-22
DE-699 | Update global server logs retrieval

aMahanna:
- announcing deprecation of `arango.database.Database.read_log`
- new `arango.database.Database.read_log_entries` method

### DIFF
--- a/arango/database.py
+++ b/arango/database.py
@@ -646,7 +646,7 @@ class Database(ApiGroup):
         :raise arango.exceptions.ServerReadLogError: If read fails.
         """
         m = "read_log() is deprecated in ArangoDB 3.8 and will be removed in a future version of the driver. Use read_log_entries() instead."  # noqa: E501
-        warn(m, DeprecationWarning, stacklevel=1)
+        warn(m, DeprecationWarning, stacklevel=2)
 
         params = dict()
         if upto is not None:
--- a/arango/database.py
+++ b/arango/database.py
@@ -645,6 +645,9 @@ class Database(ApiGroup):
         :rtype: dict
         :raise arango.exceptions.ServerReadLogError: If read fails.
         """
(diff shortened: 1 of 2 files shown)


arango/database.py
```diff
@@ -615,7 +615,10 @@ def read_log(
         search: Optional[str] = None,
         sort: Optional[str] = None,
     ) -> Result[Json]:
-        """Read the global log from server.
+        """Read the global log from server. This method is deprecated
+            in ArangoDB 3.8 and will be removed in a future version
+            of the driver. Use :func:`arango.database.Database.read_log_entries`
+            instead.
```
aMahanna, line 621: Happy to adjust the way we address this for a better solution. We could for example use the `@deprecated` decorator if we add the [Deprecated](https://pypi.org/project/Deprecated/) package as a a dependency

apetenchea: I think having the comment is great, but we should make it visible for users that are not going through the documentation. Or, for example, if they use another package that is built on top of `python-arango`, they should be able to figure out that there is a problem underneath.
Using the `@deprecated` decorator would do just fine. I'm not familiar with it, but a requirement that I find sensible is to allow to user to disable it the warning.
A slightly different approach that I used before is to emit a warning using the standard library, for example [here](https://github.com/ArangoDB-Community/python-arango/blob/main/arango/database.py#L2766). [FutureWarning ](https://github.com/ArangoDB-Community/python-arango/blob/main/arango/database.py#L2766) or [DeprecatedWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) would be fit the bill (the latter is ignored by default).

apetenchea (commented):
Change looks good. We should take a moment to decide upon emitting some kind of warning, either using a decorator as you pointed out, or standard `warn`.

arango/database.py
aMahanna: makes sense!

I've added a `warn(..., DeprecationWarning)` call in dd7383a

PR #302, opened 2023-11-24
Migrate to `pyproject.toml`

aMahanna:
Migrating from storing package configuration in `setup.py`  to `pyproject.toml`. Updates `pypi.yaml` to use `python -m build` instead of `python setup.py ...`, as the latter is no longer the standard.

- https://peps.python.org/pep-0518/
- https://peps.python.org/pep-0621/
- [Why you shouldn't invoke setup.py directly](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html)


apetenchea (approved):
LGTM

PR #308, opened 2023-12-29
new: `put /_admin/compact`

aMahanna:


tests/test_database.py
```diff
@@ -263,6 +269,23 @@ def test_database_misc_methods(sys_db, db, bad_db, cluster):
...
+
+    collection = db.create_collection(generate_col_name())
+    collection.insert({"foo": "bar"})
+
+    token = generate_jwt(secret)
+    db_superuser = client.db(db.name, superuser_token=token)
+    result = db_superuser.compact()
+
+    # TODO: Why is `result` == {} ?
+    # assert result == ...
```
aMahanna, line 287: todo: figure out what to do with the assertion

apetenchea: This is ok, there's no value that I know of which should be returned. In case the server cannot honor the request, it returns 401 and the `compact()` method raises, just as above at line 273.
Is there any additional output that you're expecting?

apetenchea (approved):
LGTM

tests/test_database.py
aMahanna: Looking back now I believe I assumed similar output to the `StandardCollection.compact()` test: 

https://github.com/ArangoDB-Community/python-arango/blob/main/tests/test_collection.py#L161-L162

Happy to clean this up if we're okay with `StandardDatabase.compact()` being empty

apetenchea: These are different routes apparently. Compacting a collection is supposed to return some information.

As for compacting a whole database, there is nothing in the docs regarding the return value, so an empty one is what we're looking for: https://docs.arangodb.com/3.11/develop/http-api/administration/#compact-all-databases

aMahanna: sounds good thanks for double checking, added in c5c4a33

PR #310, opened 2023-12-29
new: `post /_admin/echo`

aMahanna:


arango/database.py
```diff
@@ -543,6 +543,27 @@ def response_handler(resp: Response) -> Json:
 
         return self._execute(request, response_handler)
 
+    def echo_request(self, body: Any) -> Result[Json]:
```
aMahanna, line 546: Curious if it would be better to just improve on `echo()` instead of introducing `echo_request`

We could have `echo()` accept an optional parameter such that if provided, will switch the `GET /_admin/echo` to a `POST`

apetenchea: I think this idea is better. The presence of the `body` parameter, for example, could be used to differentiate between the two. The rationale being, since the endpoints are very similar, it might be easier to maintain a single method with a simple if/else logic in the future, instead of two separate methods.

aMahanna: understood! will do

aMahanna: fixed in 5d272f4, let me know what you think

apetenchea (approved):
Nice, lgtm!

PR #316, opened 2024-01-16
DE-743 | MDI/ZKD Indexes

aMahanna:
Internal discussions have established that an `mdi` index is an alias for a `zkd` index, similar to the `hash` index & `persistent` index situation.

Given that python-arango never had support for `zkd` index creation, this PR introduces both `add_zkd_index` and `add_mdi_index`

i.e
- MDI (>=3.12)
- ZKD (<3.12)

Ref: https://docs.arangodb.com/3.12/index-and-search/indexing/working-with-indexes/multi-dimensional-indexes/

Happy to address this in another way, perhaps there's no need for 2 methods?

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -134,7 +134,7 @@ def pytest_configure(config):
     global_data.username = username
     global_data.password = password
     global_data.db_name = tst_db_name
-    global_data.db_version = version.parse(db_version)
+    global_data.db_version = version.parse(db_version.split("-")[0])
     global_data.sys_db = sys_db
     global_data.tst_db = tst_db
     global_data.bad_db = bad_db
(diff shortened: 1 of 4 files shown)


aMahanna:
revisiting this now, I wonder if we would benefit from just having a public `add_index()` method instead of individually defining index create methods.

Similar to [create_analyzer](https://github.com/ArangoDB-Community/python-arango/blob/main/arango/database.py#L2608-L2614) or [create_view](https://github.com/ArangoDB-Community/python-arango/blob/main/arango/database.py#L2360-L2362)

tests/conftest.py
```diff
@@ -134,7 +134,7 @@ def pytest_configure(config):
     global_data.username = username
     global_data.password = password
     global_data.db_name = tst_db_name
-    global_data.db_version = version.parse(db_version)
+    global_data.db_version = version.parse(db_version.split("-")[0])
```
aMahanna, line 137: just want to highlight this `split`. Let's us test with a nightly build like `3.12.0-NIGHTLY.20240116`

apetenchea (approved):
LGTM

Issue #234, opened 2022-12-02
Documentation issue: ordered/geoJSON on collection.add_geo_index


Adrian-at-CrimsonAzure:
In the base collection, it seems like the `ordered` index value is incorrectly documented. According to the[ArangoDB docs](https://www.arangodb.com/docs/stable/indexing-geo.html) this `geoJSON` being set to true turns on geoJSON support/indexing. Here in the docstring nothing of the sort is mentioned, and it implies that it has something to do with coordinate ordering.
https://github.com/ArangoDB-Community/python-arango/blob/86411ef40dc22b51a21eccdce8cc179b46d322f1/arango/collection.py#L1169-L1200

It took me way too long to figure out how to turn on geoJSON support because of this.

joowani:
Thank you for the report @Adrian-at-CrimsonAuzre. Will take a look when my time allows.

PR #318, opened 2024-01-19
breaking: fix `geoJson` parameter

aMahanna:
Closes https://github.com/ArangoDB-Community/python-arango/issues/234

Not sure how else to handle this

We are currently using the parameter name `ordered` to specify whether a Geo Index should be created using GeoJson data-format or not, which can be confusing to users.

This PR just renames the parameter to `geo_json`

Ref: https://docs.arangodb.com/3.11/index-and-search/indexing/working-with-indexes/geo-spatial-indexes/

### DIFF
--- a/arango/collection.py
+++ b/arango/collection.py
@@ -1385,7 +1385,9 @@ class Collection(ApiGroup):
             with at least two floats. Documents with missing fields or invalid
             values are excluded.
         :type fields: str | [str]
-        :param geo_json: Whether to use GeoJSON data-format or not.
+        :param geo_json: Whether to use GeoJSON data-format or not. This
+            parameter has been renamed from `ordered`. See Github Issue
+            #234 for more details.
         :type geo_json: bool | None
         :param name: Optional name for the index.
         :type name: str | None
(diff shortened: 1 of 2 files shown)


apetenchea (approved):
I agree with this change. Unfortunately there's no easy way around it. We'll have to state it clearly in the release notes.
It's worth doing a major release anyway, since we're adding support for multiple features.
As a user, at least the fix should be pretty obvious from the documentation.

arango/collection.py
```diff
@@ -1385,8 +1385,8 @@ def add_geo_index(
             with at least two floats. Documents with missing fields or invalid
             values are excluded.
         :type fields: str | [str]
-        :param ordered: Whether the order is longitude, then latitude.
-        :type ordered: bool | None
+        :param geo_json: Whether to use GeoJSON data-format or not.
```
apetenchea, line 1388: Might be helpful to mention the change in the docs, so that users may spot the issue easily.
```
Whether to use GeoJSON data-format or not. This parameter has been renamed from `ordered`.
```

aMahanna: good call, added in [5b0d0ae](https://github.com/ArangoDB-Community/python-arango/pull/318/commits/5b0d0aec9555231c5a8428fa6cac5fbc69209d88)

apetenchea (approved):

PR #321, opened 2024-02-02
Make `add_index` public 

aMahanna:
This PR proposes to switch `_add_index` to `add_index`, and to deprecate the `add_*_index` methods in favour of generalizing to `add_index` (similar to how we have `create_view`, `create_analayzer`, etc.)

**However**,  we need to decide if:
- The `data` parameter of `add_index` should expect `camelCase` or `snake_case` fields

The current `add_*_index` methods accept `snake_case`, convert to `camelCase` for API purposes, then output `snake_case`. We need to decide if we want to follow suit for `add_index`, or if we want to drop this whole `snake_case` to `camelCase` to `snake_case` conversion.


aMahanna:
hey @apetenchea could I get your thoughts on this (PR description) ^

wondering what the best way to proceed is

apetenchea (commented):
@aMahanna 

Since the opportunity is now there, let's drop the `snake_case` to `camelCase` conversion for the new `add_index`. This is not an easy decision, as it may seem inconvenient at first to some users. But, because this release changes the API anyways, let's take full advantage of it.

My rationale is, if we skip the conversion and just pass the arguments as they are, any additional index parameters that is going to be added in future ArangoDB releases will be supported automatically by our driver. Having that extra bit of validation is nice, but the HTTP API and the documentation are already so simple that I'm afraid the entire `snake_case` thing creates confusion sometimes. It's the pythonic way, I know, but our driver is really a convenient wrapper on top of the requests API.

In the end, our users want to use the database from Python. Let's also specify in the docstring of `data` that it's supposed to be used exactly as in the official ArangoDB HTTP docs (eg camelCase).

What do you think?

aMahanna:
@apetenchea 

I can get behind this change, but it raises the question of what should we do with [formatter.py](https://github.com/ArangoDB-Community/python-arango/blob/main/arango/formatter.py), as that is what is currently driving the `snake_case` to `camelCase` conversion logic.

The way I see it, we have two "`snake_case` vs `camelCase`" topics;
1. Should python-arango parameters of type `dict` contain keys that are `snake_case` or `camelCase`? i.e this would apply to more than just the `add_index` method
2. Should python-arango parameters of methods that interact with the ArangoDB API be `snake_case` or `camelCase`? e.g `requestTimeout` vs `request_timeout`

I apologize for widening the scope, but it's a thought that just occurred to me now after reading your proposal

Happy to move this convo somewhere else if preferred

apetenchea (commented):
@aMahanna 
You're right - this issue largely stems from the differing conventions between Python (predominantly `snake_case`) and JavaScript (`camelCase`), not to mention our primary use of `camelCase` in the ArangoDB C++ codebase. Given that the Python driver was initially a community-led project adhering to PEP8 guidelines, the prevalence of `snake_case` parameters among our Python users is understandable. Some IDEs would even emit a small warning when `camelCase` is used for parameter names.

For dictionary parameters, aligning with camelCase would ideally mirror the 1-1 mapping with our HTTP API, reducing confusion. However, the practicality of retroactively changing this across all methods poses a significant challenge, particularly in terms of user inconvenience and the potential for breaking backward compatibility. While achieving uniformity across all dict-based parameters is conceptually appealing, the implications for our users' existing codebases cannot be understated.

Regarding non-dict parameters, the impact of transitioning to camelCase would be even more profound, necessitating widespread changes on the user's end. A gradual approach, initially supporting both `snake_case` and `camelCase` conventions while providing ample warning, and then eventually switching off the legacy one, might mitigate backlash, but this is not without its own complexities. 

I propose a cautious approach: eliminate `snake_case` conversion for the newly added `add_index` method while maintaining current conventions for legacy methods, using the formatter upon return, but just for the methods that were already there, such as `add_hash_index`. The potential UX inconsistency that adopting `camelCase` for just the add_index method might introduce makes me scratch my head too (see [Database::create_view](https://github.com/ArangoDB-Community/python-arango/blob/main/arango/database.py#L2545) for example). I think such choices were made under constraints like development time, which goes to show that indeed things are much simpler when sticking to the same convention as the database API (in our case, camel).

aMahanna:
@alexbakharew 

agreed! moving forward with this will give us a chance to get community feedback

apetenchea:
The `add_index` takes a `data` parameter - the keys are pure ArangoDB REST API arguments, no `snake_case` conversion anymore. However, for the output, the legacy methods remain the same, while the new `add_index` method by default skips formatting.

I have added a generous note in the documentation explaining that. Hopefully, there should be no confusion among the users, as they switch to the `add_index` method.

arango/collection.py
```diff
@@ -1259,11 +1260,27 @@ def response_handler(resp: Response) -> Json:
...
-    def _add_index(self, data: Json) -> Result[Json]:
-        """Helper method for creating a new index.
+    def add_index(self, data: Json, formatter: bool = False) -> Result[Json]:
+        """Create an index.
+
+        .. note::
 
-        :param data: Index data.
+            As the `add_index` method was made available starting with driver
+            version 8, we have decided to depreciate the other `add_*_index`
```
aMahanna, line 1269:
```suggestion
            version 8, we have decided to deprecate the other `add_*_index`
```

arango/collection.py
```diff
@@ -1369,6 +1387,9 @@ def add_skiplist_index(
         :rtype: dict
         :raise arango.exceptions.IndexCreateError: If create fails.
         """
+        m = "add_skiplist_index is deprecated. Using add_index with {'type': 'persistent'} instead."  # noqa: E501
```
aMahanna, line 1390:
```suggestion
        m = "add_skiplist_index is deprecated. Using add_index with {'type': 'skiplist'} instead."  # noqa: E501
```

aMahanna (commented):
LGTM! just two small doc fixes

PR #322, opened 2024-02-02
DE-752 | `multi_delimiter` analyzer [3.12]

aMahanna:
Can be merged after `3.12` image is introduced

tested to work locally with `3.12.0-NIGHTLY.20240305`


tests/test_analyzer.py
```diff
@@ -71,3 +81,16 @@ def test_analyzer_management(db, bad_db, cluster, enterprise, db_version):
             "format": "latLngDouble",
         }
         assert db.delete_analyzer(analyzer_name)
+
+    # Test create delimieter analyzer with multiple delimiters
+    # NOTE: This is currently failing on 3.12.0-NIGHTLY.20240116
```
aMahanna, line 86: highlighting

tests/test_analyzer.py
```diff
@@ -71,3 +81,16 @@ def test_analyzer_management(db, bad_db, cluster, enterprise, db_version):
...
+    if db_version >= version.parse("3.12.0"):
+        result = db.create_analyzer(
+            name=generate_analyzer_name(),
+            analyzer_type="delimiter",
```
alexbakharew, line 90: The analyzer type here is wrong. It should be `multi_delimiter`.
Here is the reference: https://docs.arangodb.com/3.12/index-and-search/analyzers/#multi_delimiter

aMahanna: thanks fixed in b48e105

tests/test_analyzer.py
aMahanna: resolved after testing with `3.12.0-NIGHTLY.20240305`

apetenchea (approved):
LGTM

PR #327, opened 2024-02-23
[3.12]: External versioning support

alexbakharew:
This PR adds support for External versioning which is introduced in 3.12: https://github.com/arangodb/arangodb/pull/20382

https://arangodb.atlassian.net/browse/ES-1771?atlOrigin=eyJpIjoiMzM4NDI1MDkxMzBlNDA2MjhkZjUyOGI2NDc3NmJlNzIiLCJwIjoiamlyYS1zbGFjay1pbnQifQ

https://arangodb.atlassian.net/browse/ES-1771

### DIFF
--- a/arango/collection.py
+++ b/arango/collection.py
@@ -1816,8 +1816,6 @@ class Collection(ApiGroup):
         param version_attribute: support for simple external versioning to
             document operations.
         :type version_attribute: str
-        :return: Document metadata (e.g. document key, revision) or True if
-            parameter **silent** was set to True.
         :return: List of document metadata (e.g. document keys, revisions) and
             any exception, or True if parameter **silent** was set to True.
         :rtype: [dict | ArangoServerError] | bool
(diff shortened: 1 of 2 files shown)


arango/collection.py
```diff
@@ -1812,6 +1813,11 @@ def insert_many(
             index caches if document insertions affect the edge index or
             cache-enabled persistent indexes.
         :type refill_index_caches: bool | None
+        param version_attribute: support for simple external versioning to
+            document operations.
+        :type version_attribute: str
+        :return: Document metadata (e.g. document key, revision) or True if
+            parameter **silent** was set to True.
```
apetenchea, line 1820: This duplicates the previous `:return:` docstring, which was already there. In order to keep a uniform `:return:` string among these functions, we should remove this and leave the already existing one there.

apetenchea (changes requested):
Thanks Alexey!

Change looks good, apart from the minor docstring issue. To make sure we got this parameter covered, would you please add a testcase in `tests/test_document.py`? For example, something similar to the queries given in Jan's PR. The test should be conditioned on version >= 3.12, see [test_pregel](https://github.com/arangodb/python-arango/blob/main/tests/test_pregel.py#L15) on how to do that.

apetenchea (approved):
Pefect, thanks!

PR #328, opened 2024-02-26
[DE-716] Compression of requests and responses

apetenchea:
Adding compression of requests and responses, as implemented in https://github.com/arangodb/arangodb/pull/20128

By default compression is turned off, so this change has no effect unless the user specifies one of the newly introduced parameters.

The `ArangoClient` gets two optional parameters:
- `request_compression` - Represents the compression strategy, `None` by default. The reason for passing in an object instead of a string (such as *deflate*) is to allow the client to fine tune the compression algorithm.
- `response_compression` - The equivalent of an `Accept-Encoding` string. This has no effect unless the server is configured to use encoding.

```python
client = ArangoClient(
    hosts="http://127.0.0.1:8529",
    request_compression=DeflateRequestCompression(),
    response_compression="deflate",
)
```

A new documentation page has been added specifically for guiding users on how to use compression.

### DIFF
--- a/arango/client.py
+++ b/arango/client.py
@@ -91,10 +91,10 @@ class ArangoClient:
        int: Timeout value in seconds.
     :type request_timeout: int | float
     :param request_compression: Will compress requests to the server according to
-        the given algorithm. Currently only `deflate` is supported.
+        the given algorithm. No compression happens by default.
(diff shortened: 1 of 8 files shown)


aMahanna (approved):
lgtm!

PR #331, opened 2024-03-11
new: `StandardDatabase.fetch_transaction` (#329)

### LINKED ISSUE #329: feature request: Continue an existing transaction
I've come across a case where a transaction needs to be shared across multiple systems. If we wrap the REST API we can easily achieve this by setting the `x-arango-trx-id` header. However, we would like to be able receive transaction IDs on both ends and continue the transaction seamlessly using the python-arango interface, instead of crudely performing raw queries against `/_cursor`.

I've come up with the following hack, which _does_ work, but given that `_executor` is private, and `_executor.id` specifically doesn't have a setter, I'm guessing there may be a reason it's discouraged:

```python
from arango.database import TransactionDatabase
from arango.client import ArangoClient

def continue_transaction(db: StandardDatabase, transaction_id):
    trx = TransactionDatabase(connection=deepcopy(db.conn))
    trx._executor._id = transaction_id
    return trx

db = ArangoClient(...).db(...)
trx = continue_transaction(db=db, transaction_id="1234")
trx.collection("vertex").insert({"_key": "test"})
trx.commit_transaction() # Alternatively, don't commit here and let the client who provided the transaction commit it themselves.
```

Would it make sense to support something like this directly? It seems to me like a reasonable use-case. If so, I'm happy to take a stab at developing a PR for this myself.

Moortiii:
Resolves #329.

Introduces a new method `fetch_transaction` on StandardDatabase that allows users to create a TransactionDatabase based on an existing Transaction, instead of starting a new transaction.


arango/executor.py
```diff
@@ -275,19 +279,34 @@ def __init__(
...
             data["maxTransactionSize"] = max_size
 
-        request = Request(
-            method="post",
-            endpoint="/_api/transaction/begin",
-            data=data,
-            headers={"x-arango-allow-dirty-read": "true"} if allow_dirty_read else None,
-        )
-        resp = self._conn.send_request(request)
+        self._id: str = ""
```
apetenchea, line 282: Nitpicking: there is no need to "declare" variables in Python. This line can be safely removed.

arango/executor.py
```diff
@@ -275,19 +279,34 @@ def __init__(
...
+            self._id = result["id"]
+        else:
+            request = Request(
+                method="get",
+                endpoint=f"/_api/transaction/{transaction_id}",
+            )
+            resp = self._conn.send_request(request)
+
+            if not resp.is_success:
+                raise TransactionInitError(resp, request)
```
apetenchea, line 307: I'm not convinced that it's a good idea to do this explicitly instead of calling `status()`, as the code is almost the same.

I imagine something along the lines of this:

```suggestion
            self._id = transaction_id
            self.status()
```

It's shorter and it shows your intention clearly - you do this stunt in order to check the transaction's status. If you feel like it's too short, just add a brief comment.

Note that `status()` throws `TransactionStatusError` instead of `TransactionInitError`. I have mixed feelings about this - you can choose to leave it like that, or catch it and re-throw a `TransactionInitError`, or even introduce a `TransactionFetchError`. I think all have pros/cons, depending on the user's interpretation. Feel free to do as you like regarding the throw (don't forget to update the test if you change the error).

tests/test_transaction.py
```diff
@@ -117,6 +117,37 @@ def test_transaction_commit(db, col, docs):
...
+        lock_timeout=1000,
+        max_size=10000,
+    )
+    txn_col = original_txn.collection(col.name)
+
+    assert "_rev" in txn_col.insert(docs[0])
+    assert "_rev" in txn_col.delete(docs[0])
+
+    txn_db = db.fetch_transaction(transaction_id=original_txn.transaction_id)
+
```
apetenchea, line 136: Nice test! I suggest adding another explicit check to make sure the transaction_id property stays the same.

```python
assert original_txn.id == txn_db.id
```

apetenchea (changes requested):
Thanks for the prompt followup!
Looks good overall and the logic is correct (also docs example is nice). The only thing which should be addressed is in `executor.py`, but that's merely a simplification.

arango/executor.py
Moortiii: I agree, but `mypy` disagrees. This triggers a `mypy: no-redef` error, because `self._id` is declared with a type hint in both branches of the conditional. There is an issue for this here: https://github.com/python/mypy/issues/14534 but it appears to be something they don't want to "fix".

arango/executor.py
Moortiii: Oh yeah, that makes a lot more sense. I just didn't think of doing this!

Moortiii: I think raising an explicit exception is the way to go. Is there any reason to raise it from the original exception, or do you think it's fine to just pass along the relevant parameters to our new exception?

tests/test_transaction.py
Moortiii: Agreed! Done.

arango/executor.py
apetenchea: This is fine

docs/transaction.rst
```diff
@@ -68,6 +68,15 @@ logical unit of work (ACID compliant).
     assert '_rev' in txn_col.insert({'_key': 'Lily'})
     assert len(txn_col) == 6
 
+    # Fetch an existing transaction. Useful if you have received a Transaction ID
+    # from some other part of your system or an external system.
+    original_txn = db.begin_transaction(write='students')
+    txn_col = original_txn.collection('students')
+    assert '_rev' in txn_col.insert({'_key': 'Abby'})
```
apetenchea, line 75:
```suggestion
    assert '_rev' in txn_col.insert({'_key': 'Chip'})
```

docs/transaction.rst
```diff
@@ -68,6 +68,15 @@ logical unit of work (ACID compliant).
...
+    txn_db = db.fetch_transaction(original_txn.transaction_id)
+    txn_col = txn_db.collection('students')
+    assert '_rev' in txn_col.insert({'_key': 'Mike'})
```
apetenchea, line 78:
```suggestion
    assert '_rev' in txn_col.insert({'_key': 'Alya'})
```

apetenchea (changes requested):
Change looks good to me!

Unfortunately the docs CI is complaining of duplicate keys. Just use some different names in the example.

docs/transaction.rst
Moortiii: Ahh, no issue. Done!

Moortiii:
I see that there are some failing checks still, but they don't seem to be related to anything I've touched or something that should have been impacted by my changes.

apetenchea (approved):
The failing docs test is due to the fact that Pregel has been deprecated recently. That's an update we'll have to make on our end.

Thank you for your contribution!

PR #335, opened 2024-04-19
Traversal API no longer available in 3.12

apetenchea:
The `api/traversal` endpoint is no longer available starting with ArangoDB 3.12. This renders the `arango.graph.Graph.traverse()` unusable, causing the graph doctests to fail.

This PR makes the following adjustments:
- emit a warning if traverse() is used (should still work 3.11)
- skipping traversal test on newer ArangoDB versions (3.12+)
- change the graph traversal example, such that we no longer encourage people to use it
- changed readme traversal example


apetenchea:
"latest" image issues are unrelated, these are due to ArangoDB 3.12 change.

aMahanna (approved):

PR #339, opened 2024-04-24
Config file selection

apetenchea:
The selection works as follows

- "latest" => 3.12
- full version specified, using two dots, say "3.12.0" => 3.12
- otherwise, assuming we're using only one do, say "3.12" => 3.12


aMahanna (approved):
thank you!

PR #341, opened 2024-04-25
Fix collection test

apetenchea:
`test_collection.py::test_collection_management` now fails due to `arango.exceptions.CollectionCreateError: [HTTP 400][ERR 10] Value cannot be empty. on attribute distributeShardsLike`. This happens because the `shard_like` parameter is empty. Truth be told, that was never a valid configuration, but since 3.12 the validation got better and now we're actually throwing an exception.

This PR removes the empty value and adds proper testing for the `shard_like` parameter.

With that, all *cluster* tests should turn green.


aMahanna (approved):
nice good catch!

PR #344, opened 2024-06-05
DE-819 | bring back `REQUEST_TIMEOUT` as a class attribute

aMahanna:
[Optional PR]

Unfortunately [this](https://github.com/arangodb/python-arango/issues/170#issuecomment-886861778) is no longer supported, but we can achieve something similar with this PR (see the added test cases)

Current ways that we support setting `request_timeout`:

```py
# 1
client_no_timeout = ArangoClient(request_timeout=None)

# 2
client_no_timeout = ArangoClient(http_client=DefaultHTTPClient(request_timeout=None))

# 3
class NoTimeoutHTTPClient(DefaultHTTPClient):
    REQUEST_TIMEOUT = None

    def __init__(self, *args, **kwargs):
        super().__init__(request_timeout=self.REQUEST_TIMEOUT, *args, **kwargs)

client_no_timeout = ArangoClient(http_client=NoTimeoutHTTPClient())
```

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -110,6 +110,31 @@ def test_client_http_client_attributes(db, username, password):
     assert http_client.request_timeout == 80
     assert client.request_timeout == http_client.request_timeout
 
+    client_no_timeout = ArangoClient(
+        hosts="http://127.0.0.1:8529", request_timeout=None
+    )
+
+    assert client_no_timeout.request_timeout is None
+
+    client_no_timeout = ArangoClient(
+        hosts="http://127.0.0.1:8529",
+        http_client=DefaultHTTPClient(request_timeout=None),
+    )
+
+    assert client_no_timeout.request_timeout is None
+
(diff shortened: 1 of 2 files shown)


tests/test_client.py
```diff
@@ -110,6 +110,31 @@ def test_client_http_client_attributes(db, username, password):
...
+        http_client=DefaultHTTPClient(request_timeout=None),
+    )
+
+    assert client_no_timeout.request_timeout is None
+
+    class NoTimeoutHTTPClient(DefaultHTTPClient):
+        REQUEST_TIMEOUT = None
+
+        def __init__(self, *args, **kwargs):
+            super().__init__(request_timeout=self.REQUEST_TIMEOUT, *args, **kwargs)
```
aMahanna, line 130: previously, the `NoTimeoutHTTPClient` with a class `REQUEST_TIMEOUT` would work without needing an `__init__`. This is PR basically introduces the ability to support `REQUEST_TIMEOUT` again, but with the exception of needing an `__init__`

PR #346, opened 2024-07-17
RTD update

apetenchea:
Removal of Sphinx context injection at build time has little effect for us.
Read [the blog post](https://about.readthedocs.com/blog/2024/07/addons-by-default/) from Read The Docs for more information.

However, this PR also updates the way docs are being build:
- We're using RTD build hook, which adds a new check for every PR (named docs/readthedocs.org:python-driver-for-arangodb). This comes with the park that we can now easily view the documentation build for every PR by clicking on details (a fake documentation page is generated).
- The docs action is not needed to build the documentation anymore. However, we'll keep it around for now in order to run the docs tests.

### DIFF
--- a/arango/request.py
+++ b/arango/request.py
@@ -12,7 +12,7 @@ def normalize_headers(
     if driver_flags is not None:
         for flag in driver_flags:
             flags = flags + flag + ";"
-    driver_version = "9.0.0"
+    driver_version = "8.1.0"
     driver_header = "python-arango/" + driver_version + " (" + flags + ")"
     normalized_headers: Headers = {
         "charset": "utf-8",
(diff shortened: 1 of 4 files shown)


arango/request.py
```diff
@@ -12,7 +12,7 @@ def normalize_headers(
     if driver_flags is not None:
         for flag in driver_flags:
             flags = flags + flag + ";"
-    driver_version = "7.7.0"
+    driver_version = "9.0.0"
```
apetenchea, line 15: This is not related, but we need to update it before the upcoming release.

aMahanna: Am I understanding correctly that this PR warrants another major release? 

Is there a way we could keep the driver to `8.x.x` for these changes?

aMahanna (approved):
looks good! just left one small question

arango/request.py
apetenchea: It's not particularly this PR , but this one: https://github.com/arangodb/python-arango/pull/192. 
But you're right, we could keep the driver to 8.x.x, it's not that big of a deal.

PR #348, opened 2024-09-08
Integrate latest 3.12.2 changes

apetenchea:
* Add API GET `/_admin/cluster/vpackSortMigration/status` to query the status of the vpack sorting migration on dbservers, single servers and coordinators.
* The `_admin/log/level` endpoint has also been extended to with an optional
  parameter `withAppenders`.
* There is also a migration API under GET `/_admin/cluster/vpackSortMigration/check` and PUT `/_admin/cluster/vpackSortMigration/migrate` to check for problematic indexes and - provided there are none - to migrate the instance to the new sorting order.

### DIFF
--- a/arango/cluster.py
+++ b/arango/cluster.py
@@ -15,6 +15,7 @@ from arango.exceptions import (
     ClusterServerRoleError,
     ClusterServerStatisticsError,
     ClusterServerVersionError,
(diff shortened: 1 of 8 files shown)


aMahanna (approved):
lgtm!

PR #349, opened 2024-09-30
update: `create_graph` & `properties` on 

aMahanna:
- Add `satellite_collections` parameter to `create_graph()`
- docstring fix in `create_graph()`
- Return `disjoint` in `properties()`


apetenchea (approved):
LGTM

PR #355, opened 2024-12-06
ES-2276 | support level 2 query profiling

aMahanna:
- Introduces support for setting the `profile` parameter in `db.aql.execute` to `2`, which returns additional `nodes` information via `cursor.statistics()`, along with the query execution plan (newly available via `cursor.plan()`)
- Warrants an `8.2.0` release

```
pip install git+https://github.com/arangodb/python-arango@ES-2276
```

```py
from arango import ArangoClient

db = ArangoClient().db()

cursor_a = db.aql.execute('FOR d IN doc SORT d._key RETURN d', profile=True)
cursor_b = db.aql.execute('FOR d IN doc SORT d._key RETURN d', profile=1)
cursor_c = db.aql.execute('FOR d IN doc SORT d._key RETURN d', profile=2)

assert cursor_a.statistics().keys() == cursor_b.statistics().keys()
assert cursor_a.profile().keys() == cursor_b.profile().keys()
assert cursor_a.plan() == cursor_b.plan() == None
assert "nodes" not in cursor_a.statistics()

assert cursor_c.plan().keys() == {'nodes', 'estimatedNrItems', 'collections', 'rules', 'variables', 'isModificationQuery', 'estimatedCost'}
assert "nodes" in cursor_c.statistics()
```

### DIFF
--- a/arango/cursor.py
+++ b/arango/cursor.py
@@ -40,6 +40,7 @@ class Cursor:
         "_count",
         "_cached",
         "_stats",
+        "_plan",
         "_profile",
         "_warnings",
         "_has_more",
@@ -63,6 +64,7 @@ class Cursor:
         self._count: Optional[int] = None
         self._cached = None
         self._stats = None
+        self._plan = None
         self._profile = None
         self._warnings = None
         self._next_batch_id: Optional[str] = None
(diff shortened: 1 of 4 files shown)


arango/cursor.py
```diff
@@ -132,12 +134,18 @@ def _update(self, data: Json) -> Json:
                 self._warnings = extra["warnings"]
                 result["warnings"] = extra["warnings"]
 
+            if "plan" in extra:
+                self._plan = extra["plan"]
+                result["plan"] = extra["plan"]
```
aMahanna, line 139: we're starting to blur the lines between camelCase & snake_case results...

`plan` is a dictionary, with some fields (such as `nodes` or `variables`) being list of dictionaries, whose fields are not consistent among each other. 

It becomes a slippery slope trying to convert every single field within the `nodes` & `variables` dictionaries into snake_case, so I'm opting towards just returning the results as-is.

Similar conversation here: https://github.com/arangodb/python-arango/pull/321#pullrequestreview-1872544091

PR #361, opened 2025-02-23
3.12.4 updates

apetenchea:
This PR prepares a next release and adds the following:

- Support for `usePlanCache` option (since 3.12.4)
- There was also a bug in the way we used to handle the `cache` parameter. It is part of the `options` sub-object.
- Removing support for Python 3.8
- Support for auto-generation of `errno.py`.


errors.py
aMahanna, line 1: confirming, can this file be removed?

.circleci/config.yml
```diff
@@ -12,7 +12,7 @@ workflows:
           matrix:
             parameters:
               # TODO: Revisit why pyenv doesn't recognize 3.12
-              python_version: ["3.8", "3.9", "3.10", "3.11"] # "3.12"
+              python_version: ["3.9", "3.10", "3.11"] # "3.12"
```
aMahanna, line 15: On the topic of python versions, I've opened this: https://github.com/arangodb/python-arango/pull/362

Feel free to merge after this PR

errors.py
apetenchea: Addressed in f0349314ec5d0646a24fdf2dde81435322031ef3

PR #368, opened 2025-03-28
Mention async driver

apetenchea:
I was thinking to do a new release anyway, and this is a good opportunity to mention the other driver.


aMahanna (approved):
🚀

PR #372, opened 2025-06-01
API Updates

apetenchea:
This PR does not modify much of the implementation, but adds or removes parameters to some of the graph-related functions.

- Vertex and Edge collections don't (actually never did) support the `silent` parameter. This could be confusing and it just makes type hinting more awkward because we have to return an `Union` instead of the actual type. See https://docs.arangodb.com/stable/develop/http-api/graphs/named-graphs/.
- Creating a vertex or edge collection has an additional `options` parameter which was not available in the driver.
- `replace_edge_definition` and `replace_vertex` were missing the `sync` parameter.
- The `edge_collections` method was missing. It corresponds to https://docs.arangodb.com/stable/develop/http-api/graphs/named-graphs/#list-edge-collections
- Added comments in `graph.rst` in order to clear up confusion about our "Edge definitions" example (see https://github.com/arangodb/python-arango/pull/371)

Changed the driver version, so we can do a new release.

### DIFF
--- a/arango/collection.py
+++ b/arango/collection.py
@@ -3008,9 +3008,8 @@ class VertexCollection(Collection):
         self,
         vertex: Json,
         sync: Optional[bool] = None,
-        silent: bool = False,
         return_new: bool = False,
-    ) -> Result[Union[bool, Json]]:
+    ) -> Result[Json]:
         """Insert a new vertex document.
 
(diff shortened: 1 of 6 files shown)


aMahanna (approved):
Thanks! Was always wondering why we never had `edge_collections()` :)

PR #373, opened 2025-06-09
Implement getIndexes method

alexbakharew:


arango/collection.py
```diff
@@ -1299,6 +1299,34 @@ def response_handler(resp: Response) -> Json:
 
         return self._execute(request, response_handler)
 
+    def get_indexes(self, with_stats: bool, with_hidden: bool) -> Result[Json]:
```
apetenchea, line 1302: Hey @alexbakharew,

Can you add these two parameters to the `indexes()` (at line 1220) method instead?
I think having two methods will get redundant, and `indexes()` is already a known part of the API and documented.

Please either include these parameters as `Optional[bool]` and set them to `None` by default, or just `bool` with a default value of `False`. You might have  to adjust the `format_index` method in order to get the full result back.

PR #376, opened 2025-07-28
Deprecate load/unload methods

apetenchea:
The collection load and unload methods have long been deprecate and are just NOPs since 3.9.

We should at least let the users know that these methods have no effect, and can safely be removed from their codebase.

See
- https://docs.arangodb.com/stable/develop/http-api/collections/#load-a-collection
- https://docs.arangodb.com/stable/develop/http-api/collections/#unload-a-collection


aMahanna (approved):
Thanks!

